PR #133, opened 2015-01-19
Parser

### LINKED ISSUE #124: beastmaster boar/hawk ability 5580
This ability is not showing up properly in the build order, needs to be manually corrected

Might as well do oracle abilities while we're at it

howardc93:
fixes #124


parser.js
```diff
@@ -157,35 +173,47 @@ function uploadToS3(archiveName, body, cb) {
...
+                            }
+                        })
+                        job.failed().error(err);
+                        //don't retry
+                        return cb(null)
+                    }
+                    //retry
+                    return cb(err);
+                }
+                utility.runParse(fileName, function(err, output) {
```
albertcui: It's possible the match is already parsed due to user upload after we finish downloading. Should we check for that?

parser/dependency-reduced-pom.xml
```diff
@@ -27,7 +27,7 @@
             <configuration>
               <transformers>
                 <transformer>
-                  <mainClass>albert.stats.Main</mainClass>
+                  <mainClass>yasp.Main</mainClass>
```
albertcui: :(

parser/pom.xml
```diff
@@ -64,7 +63,7 @@
                         <configuration>
                             <transformers>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
-                                    <mainClass>albert.stats.Main</mainClass>
+                                    <mainClass>yasp.Main</mainClass>
```
albertcui: :(:(

parser.js
howardc93: https://github.com/albertcui/YASP/pull/133/files#diff-09461573a85e2d94f056dd6814769042R177

that's what this check does

albertcui: Yes, but you do that /before/ the download. I'm saying while downloading it could have been parsed.

howardc93: Download is the most important thing to save since it's time consuming and we have a quota.  If we already downloaded the replay might as well parse it

PR #199, opened 2015-02-03
Timeline page

albertcui:


howardc93:
Test is failing due to moved content, fix this test and add a new one for the new page.

PR #271, opened 2015-02-27
Re-enable cache

### LINKED ISSUE #258: Reenable Cache

albertcui:
Fixes #258

### DIFF
--- a/routes/matches.js
+++ b/routes/matches.js
@@ -31,11 +31,15 @@ matches.get('/', function(req, res) {
     });
 });
 matches.param('match_id', function(req, res, next, id) {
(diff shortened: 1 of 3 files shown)


routes/matches.js
```diff
@@ -31,11 +31,15 @@ matches.get('/', function(req, res) {
     });
 });
 matches.param('match_id', function(req, res, next, id) {
-    redis.get(id, function(err, reply) {
-        if (!err && reply && false) {
+    redis.get("match:" + id, function(err, reply) {
```
howardc93: optional, but define the key once since you use it twice

Issue #334, opened 2015-03-29
Show a small colored box for the player's color


evaryont:
Players will sometimes refer to another based on the color (teal, pink, brown, etc. [The chat for match 1357893108 for example](http://yasp.co/matches/1357893108/chat)) and not their name. While I think the ordering is almost always the same so I could memorize it, it would be convenient to show the player's color. Perhaps a thin border to the left of the player?

howardc93:
Would need to define mapping of slot number to hex color.  If you could do that for us, actually adding it would be pretty trivial.

howardc93:
We have to be careful of 1v1 games, though, so the mappings should probably be with player slot, which go 1,2,3,4,5,128,129,130,131,132.

The only annoying part is getting the correct hex values.  Anyone feel like tackling that?

McHearty:
Blue - #2E6AE6
Teal - #5DE6AD
Purple - #AD00AD
Yellow - #DCD90A
Orange - #E66200
Pink - #E67AB0
Pus Yellow - #92A440
Light Blue - #5CC5E0
Green - #00771F
Brown - #956000

These are all the colours used ingame

howardc93:
Thanks!  @albertcui can you get this in tonight?  Could possibly push directly to master, since it's UI only.

howardc93:
@McHearty if you'd like to get a contrib in, you could add these to sources.json for us, just
```
"player_colors":{
"1":"#2E6AE6",
...
"128":"#E67AB0",
...
}
```

McHearty:
@howardc93 added a pull request

albertcui:
Thanks! Will try to get this in tonight.

howardc93:
Cool, just need to update the Jade templates to use the colors now.  If we're feeling ambitious we could probably get c3 to use these colors as well. . .

albertcui:
This is on production right now, [here](http://yasp.co/matches/1360191875/chat) for example.

PR #336, opened 2015-03-29
#334 Added Player Colours to sources.json

McHearty:
Contrib yo


sources.json
```diff
@@ -69,6 +69,18 @@
...
+    "2": "#5DE6AD",
+    "3": "#AD00AD",
+    "4": "#DCD90A",
+    "5": "#E66200",
+    "128": "#E67AB0",
+    "129": "#92A440",
+    "130": "#5CC5E0",
+    "131": "#00771F",
+    "132": "#956000"
+  }
```
albertcui: Need a comma here:

```
},
```

howardc93:
Should we call it player_colors or player_slot_colors, or does it not really matter? :P

McHearty:
Player colours currently are only used in this way, so unless you want to future proof against any possible changes, player_colors should be sufficient

howardc93:
sounds good.  signing off

PR #337, opened 2015-03-30
April fools

### LINKED ISSUE #318: tooltips need to be re-initialized when paginating in datatables
Right now they don't work on subsequent pages since datatables redraws dom elements. . .

This is why I don't want to use datatables for everything :P

albertcui:
Fixes #318
Adds 322 Level
Improves trends page loading
Adds white theme option


_posts/v.2.0.0.md
```diff
@@ -5,20 +5,20 @@ YASP is jumping a version number, releasing v2.0.0!
 <!--more-->
 
 Features
-====
+--------
```
howardc93: Should we just go back to the previous (no heading, just a line break)

howardc93:
see comment, signoff otherwise

PR #371, opened 2015-04-20
bust cache on req

albertcui:


web.js
```diff
@@ -102,7 +102,9 @@ io.sockets.on('connection', function(socket) {
                     job.on('complete', function(result) {
                         console.log(result);
                         socket.emit('log', "Request Complete!");
-                        socket.emit('complete');
+                        redis.del("match:" + match_id, function(err, resp) {
```
howardc93: error should be handled

PR #447, opened 2015-05-31
added env var to optionally disable the captcha on the Request page

nickhh:
useful if you haven't been captcha whitelisted

### DIFF
--- a/config.js
+++ b/config.js
@@ -26,6 +26,7 @@ var defaults = {
     "PROXY_URLS": "", //comma separated list of proxy urls to use
     "STEAM_API_HOST": "api.steampowered.com",
     "ROLE": "", //for specifying a node type
+    "DISABLE_RECAPTCHA": "false", // if "false" the recaptcha on the Request page will be enabled, otherwise it will be disabled
     //the following are deprecated
     "AWS_S3_BUCKET": "",
     "AWS_ACCESS_KEY_ID": "",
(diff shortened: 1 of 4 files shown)


config.js
```diff
@@ -26,6 +26,7 @@ var defaults = {
     "PROXY_URLS": "", //comma separated list of proxy urls to use
     "STEAM_API_HOST": "api.steampowered.com",
     "ROLE": "", //for specifying a node type
+    "DISABLE_RECAPTCHA": "false", // if "false" the recaptcha on the Request page will be enabled, otherwise it will be disabled
```
howardc93: set this to empty string for default, use a truthy value to enable it

socket.js
```diff
@@ -32,11 +32,12 @@ io.sockets.on('connection', function(socket) {
             var match_id = data.match_id;
             match_id = Number(match_id);
             socket.emit('log', "Received request for " + match_id);
-            if (!body.success && config.NODE_ENV !== "test") {
+            if (!body.success && config.NODE_ENV !== "test"
+                    // if the DISABLE_RECAPTCHA env var has been set, ignore a bad body.success
+                    && config.DISABLE_RECAPTCHA === "false") {
```
howardc93: if you use the truthy/falsy value you should be able to just do !config.DISABLE_RECAPTCHA

socket.js
```diff
@@ -32,11 +32,12 @@ io.sockets.on('connection', function(socket) {
...
                 console.log('failed recaptcha');
                 socket.emit("err", "Recaptcha Failed!");
-            }
-            else if (!match_id) {
+            } else if (!match_id) {
```
howardc93: I think the correct check is isNaN(match_id) since NaN !== false

howardc93:
signoff

PR #450, opened 2015-06-02
Betterstreaks

### LINKED ISSUE #442: better display of streaks
(investigate the feasibility of) instead of listing the number of times each type of streak happened, move the streaks table to the teamfights tab and display (1) the heroes that were killed in that particular streak in which order at what times and (2) associate each of those kills with specific teamfights (if applicable)

nickhh:
resolves #442, also (accidentally) contains nocaptcha branch changes

### DIFF
--- a/processParse.js
+++ b/processParse.js
@@ -495,11 +495,15 @@ function runParse(job, cb) {
                             var x = intervalState[e.time][e.slot].x;
                             var y = intervalState[e.time][e.slot].y;
                             e.type = "deaths_pos";
+                            var old_key = e.key;
                             e.key = [x, y];
                             e.posData = true;
                             populate(e, tf);
                             //increment death count for this hero
                             tf.players[e.slot].deaths += 1;
+                            //return the entry to its original state
+                            e.type = "kills";
+                            e.key = old_key;
                         }
                     }
                     else if (e.type === "buyback_log") {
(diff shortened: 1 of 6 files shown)


howardc93:
signoff

processParse.js
```diff
@@ -453,110 +479,228 @@ function runParse(job, cb) {
...
+                        populate(e, tf);
+                        //get slot of target
+                        e.slot = hero_to_slot[e.key];
+                        //0 is valid value, so check for undefined
+                        if (e.slot !== undefined) {
+                            //if a hero dies, add to deaths_pos, lookup slot of the killed hero by hero name (e.key), get position from intervalstate
+                            var x = intervalState[e.time][e.slot].x;
+                            var y = intervalState[e.time][e.slot].y;
+                            e.type = "deaths_pos";
+                            var old_key = e.key;
```
howardc93: I think we should actually fix this before committing

processParse.js
```diff
@@ -453,110 +479,228 @@ function runParse(job, cb) {
...
+                // index into the killer's kills log
+                var killer_info = parsed_data.players[killer_index].multi_kill_streaks;
+                var all_streak_length = killer_info.kills.length;
+                var cur_streak_length = killer_info.kills[all_streak_length-1].length;
+
+                // record which hero this is
+                killer_info.hero_id = hero_to_id[killer];
+
+                // if this entry is a valid kill notification
+                if ((entry.type === "kills" || entry.type === "kills_log")
```
howardc93: kills_log should be sufficient, that already checked for hero+illusion previously

processParse.js
```diff
@@ -453,110 +479,228 @@ function runParse(job, cb) {
...
+
+                        // add this kill (represented by a tuple) to the killer's list of kills
+                        killer_info.kills[all_streak_length-1].push([0, 0, 0, 0]);
+                        cur_streak_length++;
+
+                        // populate the tuple with some pertinent information
+                        var kill_info = killer_info.kills[all_streak_length-1][cur_streak_length-1];
+
+                        // a) record who was killed
+                        kill_info[0] = hero_to_id[killed];
```
howardc93: I prefer to use objects rather than arrays since they're more human readable for debugging, but if you're okay with it. . .

utility.js
```diff
@@ -285,7 +285,22 @@ function getParseSchema() {
                 //"ability_trigger": {}
                 "kill_streaks": {},
                 "multi_kills": {},
-                "healing": {}
+                "healing": {},
+                "multi_kill_streaks": {     // for use with processMultiKillStreaks() in processParse.js
+                    "hero_id": "",          // the hero id of this player
+                    "kills": [[]],          // an array of kill streaks
```
howardc93: still not sure why you're seeding your arrays with [], 1, 2, etc.

views/match/match_teamfights.jade
```diff
@@ -122,6 +132,111 @@ block match_content
                     td=player.damage
                     td=player.gold_delta
                     td=player.xp_delta
+
+  if parsed_data
+    - var players = parsed_data.players
+    if players
```
howardc93: ouch this is ugly

utility.js
```diff
@@ -285,7 +285,22 @@ function getParseSchema() {
...
+                                            //     where each kill streak is an array of kills
+                                            //         where each kill is an array with
+                                            //             index 0 = the hero id of the player who was killed
+                                            //             index 1 = the multi kill id of this kill
+                                            //             index 2 = the team fight id of this kill
+                                            //             index 3 = the time of this kill
+                    "cur_multi_id": 0,      // the id of the "current" (during population) multi kill
```
howardc93: don't store temporary state in the schema, do it in processParse so this isn't added to db

processParse.js
nickhh: I'll change this. (I just didn't want to mess with your code too much.)

processParse.js
nickhh: Some of the kills happen through kill events, not kill_log events. Maybe that's because of the weird re-use of the entry happening above, I turned the type back to kill instead of kill_log? Once I refactor that, this might become safer...

processParse.js
nickhh: Oh, you mean type the 4-tuple to be {killer: ""} etc? I could do that? I'm fine with it like this, but if you prefer it the other way it's an easy change.

views/match/match_teamfights.jade
nickhh: It's super safe :P Why don't you like it?

processParse.js
howardc93: yep thats probably it

utility.js
nickhh: I'll refactor this for space efficiency.

processParse.js
howardc93: i find that more readable but youll be maintainimg this

views/match/match_teamfights.jade
howardc93: it's just ugly.  you can use && to make it look a little better

nickhh: OH you mean the lines under this comment, got it :P

utility.js
nickhh: I fixed it so it's not longer needed :P

nickhh:
I made all of Howard's suggested changes.

nickhh:
The build failed because "web player page tests /players/:valid (no matches): Error: timeout of 2000ms exceeded." Is that not my fault? (Travis seems to be super buggy :P )

howardc93:
I can re-run the build, or you can test it locally to see it fails there as
well

PR #466, opened 2015-06-09
self word cloud

nickhh:
created a second word cloud that shows only the words that YOU typed into all chat

### DIFF
--- a/advquery.js
+++ b/advquery.js
@@ -171,7 +171,7 @@ function advQuery(options, cb) {
             }
 
             // determine which user page this information is for
-            var requesting_player = parseInt(options.select["players.account_id"]) || false;
+            var requesting_player = parseInt(options.select["players.account_id"]);
 
             // console.timeEnd("parsedplayerdata");
             // console.time('compute');
@@ -394,8 +394,11 @@ function aggregator(matches, fields) {
         "kill_streaks": function(key, m, p) {
             standardAgg(key, p.parsedPlayer.kill_streaks, m);
         },
-        "word_counts": function(key, m, p) {
-            standardAgg(key, m.word_counts, m);
+        "all_word_counts": function(key, m, p) {
+            standardAgg(key, m.all_word_counts, m);
+        },
+        "my_word_counts": function(key, m, p) {
+            standardAgg(key, m.my_word_counts, m);
         }
     };
     //if null fields passed in, do all aggregations
--- a/advquery.js
+++ b/advquery.js
@@ -11,6 +11,10 @@ var isSignificant = helper.isSignificant;
 var async = require('async');
 
 function advQuery(options, cb) {
+
+    var agg_player_page = "time to aggregate player page";
+    console.time(agg_player_page);
+
     var default_project = {
         start_time: 1,
         match_id: 1,
@@ -25,8 +29,10 @@ function advQuery(options, cb) {
         radiant_name: 1,
         dire_name: 1
     };
+
     options.project = options.project || default_project;
     options.project.players = 1;
+
     //only project the fields we need
     options.project["players.account_id"] = 1;
     options.project["players.hero_id"] = 1;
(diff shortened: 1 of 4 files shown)


advquery.js
```diff
@@ -113,13 +126,16 @@ function advQuery(options, cb) {
         sort: options.sort,
         fields: options.project
     };
-    console.log(options);
-    console.time('db');
+
+    // console.log(options);
+    // console.time('db');
```
howardc93: are you removing the more specific timing breakdowns for a single time that covers everything?

compute.js
```diff
@@ -184,34 +222,52 @@ function computeMatchData(match) {
...
-                    }
-                });
-                match.players.forEach(function(player) {
-                    player.parsedPlayer.pick_order = pick_map[player.player_slot];
-                });
+
+                // aggregation of all words in all chat this player has experienced
+                match.all_word_counts = count_words(match.parsed_data.chat, null);
+                // aggregation of only the words in all chat this player said themselves
+                match.my_word_counts = count_words(match.parsed_data.chat, requesting_player);
```
howardc93: rather than pass requesting_player, players[0] should work.  I copy the players array to all_players and only put a single player in players.

compute.js
```diff
@@ -8,175 +8,213 @@ var sentiment = require('sentiment');
...
+
+                // count the words that occur in a set of messages
+                // - messages: the messages to create the counts over
+                // - player_filter: if non-null, only count that player's messages
+                function count_words (messages, player_filter) {
+
+                    // extract the message strings from the message objects
+                    var chat_words = messages.map(
+                        function(message) {
+                            return (!player_filter || match.all_players[message.slot].account_id
```
howardc93: This means that all "anonymous" players will have all messages from "anonymous" attributed to them.

You also need to beware of the case where number of players < 10.  message.slot ranges from 0 to 9 always but match.all_players can have 2-10 entries.

advquery.js
nickhh: yes. was there a reason for having them? i can put them back. the entire computation took less than a second

compute.js
nickhh: cool, i'll try it

views/player/player_wordcloud.jade
```diff
@@ -1,19 +1,34 @@
...
+      my_list.push([key, my_counts[key]]);
     }
-    console.log(list);
-    WordCloud(document.getElementById('wordcloud'), { list: list, weightFactor:5, minSize: 10 } );
+
+    for (var key in all_counts){
+      all_list.push([key, all_counts[key]]);
+    }
+
+    WordCloud(document.getElementById('my_word_counts'), { list: my_list, weightFactor:5, minSize: 10 } );
```
howardc93: you want to take a stab at rescaling the values (so the most common word is a standard size, and everything else scaled around it?)

advquery.js
howardc93: Not super important, but they were there to detect if any stage was causing a slowdown.  Usually it's the parsedplayerdata stage that can take a long time...

compute.js
nickhh: 1) player_filter is an account id integer, not a string
2) i believe this would only break is the information in message.slot was wrong

howardc93: You can take a look here:

https://github.com/yasp-dota/yasp/blob/master/views/match/match_chat.jade#L15

Adapt that approach to convert message.slot into an index into match.players and use player_slot within that to match.

views/player/player_wordcloud.jade
nickhh: sure. you just want the word cloud to look nicer?

howardc93: As it is currently sometimes words get displayed or skipped because of size issues.

For bonus points, make a list of common English words like "i", "the", "a" and remove them

compute.js
howardc93: Anonymous has an account ID, it's the max 32 bit unsigned integer.

Message.slot is computed always based on 10 players.  So yes, it could be wrong if match.players doesnt have 10 players.

advquery.js
nickhh: changed this back

compute.js
nickhh: players[0] does not work

howardc93: what's the content of match.players?

compute.js
nickhh: k, did this

views/player/player_wordcloud.jade
nickhh: i only kind of resolved this. i made all the words show up, but didn't fix the scaling issue

howardc93: how?

PR #478, opened 2015-06-23
skill/pro games page

### LINKED ISSUE #299: skill brackets
Probably not gonna be done, but I'll just document here:

Steam API only allows querying for skill on one API:
https://api.steampowered.com/IDOTA2Match_570/GetMatchHistory/V001/?key=XXXXX&start_at_match_id=1299599890&skill=3

The values for skill are 
skill=<skill>                  # 0 for any, 1 for normal, 2 for high, 3 for very high skill (default is 0)

The matches are in descending order by match_id.

However, only the first 500 results can be returned.  Since the output is sorted by match_id but matches don't show up in the API until completion, the sample is unfairly weighted toward shorter matches.

Implementation:
Essentially, we'd have to run a background job that queries the 500 available matches for each skill level 1, 2, 3.  For each match, we have to check our DB to see if we can fill in skill data.  It's not a very fun implementation.

### LINKED ISSUE #480: pro games page
since we now parse all professional matches, make a page that lists them

howardc93:
fixes #299 
fixes #480 

still need to:
- [x] add db inserting code
- [x] add to yasp-core deployment command
- [x] add to procfiles
- [x] display skill on match pages/matches table
- [x] clean up pro page ui (at a very basic level)
- [x] add league data to matches table


howardc93:
still need to:
- [ ] add db inserting code
- [ ] add to yasp-core deployment command

howardc93:
closing PR (merge with #489 )

PR #490, opened 2015-07-03
Fixed sorting recent matches by played column

coreymaher:
Sorting by the played column attempts to sort based on the the human readable "time from now" which will never work as expected.

This pull request changes sorting to use the unix offset instead.

I didn't set up YASP to actually see if this pull request works, so please let me know if there are any issues.


howardc93:
Thanks for this!  Tested and merged.  I made it return just start_time+row.duration in the case that it's a sort so we can skip the creation of a moment object in that case.

coreymaher:
Ah yeah that makes a lot more sense

PR #510, opened 2015-07-09
add shaSentryfile and authCode support

nikolaylukyanchuk:
add shaSentryfile and authCode  support on steam login (fix restart problem), save shaSentryfile on redis


howardc93:
Unfortunately our retrievers are run on VMs without redis access, so merging this as-is would break production.

We also use accounts that don't have Steam Guard to avoid this issue.

PR #526, opened 2015-07-14
quick button-based solution for histogram page - Speed Up!

Aehrraid:
(issue: https://github.com/yasp-dota/yasp/issues/460)

For every histogram graph there is a button titled with "Duration", "Gold per Minute". The button is a toggle-button (just like filter above). The button becomes darker if toggled. Initially the duration graph is shown.

I could experience a third less loading time.. probably, because most of the graphs are "display:none" and not rendered therefore. Furthermore firefox instances do not show the "stop script" dialogue, because of reduced js-execution time.

Button title processing needs some improvement though. --> awkward redundant key 2 title transformation
There is also an issue concerning the initially shown duration-chart. If it is clicked, it collapses and reopens quickly. It should close though..


views/player/player_histograms.jade
```diff
@@ -10,12 +10,18 @@ block playercontent
           var histogramData = {};
         each val, key in player.aggData
           if key in histograms
+            - var collapsed = key != "duration"
```
howardc93: should use !==

views/player/player_histograms.jade
```diff
@@ -10,12 +10,18 @@ block playercontent
...
+            button.btn.btn-xs.btn-dark.toggle-button(data-toggle="collapse" data-target="#chart-histogram-container-"+key, aria-expanded=collapsed ? "false" : "true", style="margin-right: 2px", class=collapsed ? "collapsed" : "") #{key.split("_").map(function(s){ return s.charAt(0).toUpperCase() + s.slice(1)}).join(" ")}
+        each val, key in player.aggData
+          if key in histograms
             - var avg = val.sum/val.n
-            h3 #{key.split("_").map(function(s){ return s.charAt(0).toUpperCase() + s.slice(1)}).join(" ")} 
-              small (#{val.n} matches, <span data-format-seconds=#{avg} class=#{key in times ? "format-seconds" : ""}>#{(avg).toFixed(2)}</span> mean)
-            script.
-              histogramData["#{key}"] = !{JSON.stringify(player.aggData[key])};
-            div.histogram-chart(id="chart-histogram-"+key, data-histogram=key)
+            - var collapsed = key != "duration"
```
howardc93: !==

howardc93:
Thanks for this!  We've been looking for a solution to this problem.

Comments:
* I think it needs to be more immediately clear to the user what they need to do to get the histograms to show up.
* The toggled class isn't very obvious.  It's hard to tell which ones are enabled and which are not.  Perhaps making this obvious would also fix the first issue (maybe a checkbox, or opacity?)
* Trying to toggle "Duration" immediately doesn't work (requires a second click to get it to hide)

Aehrraid:
Thanks!

Regarding your ideas:
- We could enhance the toggle-buttons with check-marks.
- We could change the css-color a bit more. Youre right. Checkboxes might solve this aswell.
- I mentioned this above. But I don't see a quick fix for this. Maybe something like:
   - Collapse all graphs.
   - Simulate a .click() on one toggle-button. (This workarounds usually result in correct behaviour)

I found another very minor issue: Toggle-buttons are not operating when using enter or space while tabbing through.

howardc93:
I think the second workaround, the default behavior is then to be closed and then you can simulate clicks for the "default" ones that should show up.

Aehrraid:
Yes I did not point it out very clearly, but you fully understood, what I mean:
1. Collapse all graphs.
2. Toggle those graphs we want to show by default. (with mentioned simulation).

I did not mean to present 'collapse all graphs' as an independend solution. ;-)

Working on it!

howardc93:
Another possibility: Show only one histogram at a time (clicking a new button switches them out)

Aehrraid:
@howardc93:
>Another possibility: Show only one histogram at a time (clicking a new button switches them out)

As far as yasp.co remains a statistics page for deep-analytics of specialized unique data, I would say that this limits the users. If I want to compare win-by-duration and win-by-xpm, I would not be able to do so. I do not think that this is a good idea. Maybe test my solution. It works quite properly.

howardc93:
I still think the toggle could be more obvious, but otherwise it looks fine.

@albertcui should review.

views/player/player_histograms.jade
```diff
@@ -10,12 +10,18 @@ block playercontent
           var histogramData = {};
         each val, key in player.aggData
           if key in histograms
+            - var collapsed = true || key !== "duration"
```
howardc93: the || is unnecessary.

In fact the entire collapsed variable is now simply true.

Aehrraid: dead code. sorry

albertcui:
Love the change. Thanks for your contribution!

PR #544, opened 2015-07-19
Caching

albertcui:


albertcui:
TODO:

[ ] Fix responsiveness of front page
[ ] Make toggle theme in footer not make the footer so much taller
[ ] Investigate table row borders

howardc93:
Light theme looks pretty bad ATM, it will either need to be removed or we'll need to work on it before deploying.

PR #550, opened 2015-07-21
[WIP] ES6 + Linting

vramana:


public/js/histograms.js
```diff
@@ -77,82 +71,79 @@ module.exports = function createHistogram(counts, win_counts, label) {
...
                         t = Number(t) / scalef;
-                        //these labels are times and need to be formatted
+                        // these labels are times and need to be formatted
                         var times = {
-                            "duration": 1,
-                            "first_blood_time": 1
+                            duration: 1,
+                            first_blood_time: 1
                         };
                         if (times[label]) {
```
vramana: @howardc93 @albertcui Isn't this reduntant code here? `times` object is has no key label clearly from the declaration.

howardc93: label is a variable.  So we may have that key.

vramana: Sorry my bad

howardc93:
How thoroughly is this tested?

vramana:
Hey, I haven't tested this. My machine is taking too long to even build. So, I didn't test this yesterday I will today. In the meantime Can you build this locally once and report any problems? I don't expect anything to break as such.

vramana:
I have tested this branch by running `npm start web` everything seems to render fine except I don't have redis database. May be you can check that part.

howardc93:
@albertcui do you want to weigh in on ES6?  Eventually we'll be able to avoid the need to use babel, right?

Does the eslint slow the build significantly? (How long does it take to run?)

vramana:
@howardc93 It doesn't slow the build at all. May be it takes around 5-10 secs to do the linting.

howardchung:
Playing a bit with React now so I've added babel to the build step and can now use ES6.  Unfortunately the code has changed a LOT since the last commit.

I'm still hoping eventually we won't need babel anymore.

howardchung:
Please see https://github.com/yasp-dota/ui if you'd like to help contribute to the new UI.  We're using ES6 for that.

PR #551, opened 2015-07-21
fix for issue 527 (toggle-button color and style)

### LINKED ISSUE #527: Toggle-button color
With my commits (https://github.com/yasp-dota/yasp/pull/526) a new minor issue appeared. Every toggle-button is darker, if collapsed. This is accomplished with:
```css
.toggle-button.collapsed {
  background-color...
}
```
**The problem:**
Every existing toggle-button (especially the Filter-buttons on every page) are not affected by this css-rule, because the .jade-source only states 'data-toggle="collapse"', not class="collapsed".
**The result:** (except player-histograms)
Initially toggle-buttons are not effected by css. They are light. But after uncollapsing and collapsing they become dark. Which is the correct state now.

Aehrraid:
issue: https://github.com/yasp-dota/yasp/issues/527

fixes #527


howardc93:
This looks good to me but I'll let @albertcui take a look first.

Aehrraid:
Correct. Thanks alot!

Aehrraid:
What is the status about this?

howardc93:
waiting for @albertcui to sign off and merge

Aehrraid:
Added another tooltip improvement. How can I assign this change to a new pull request? (not very firm with git(hub))

howardc93:
This is a separate change, so it should be made a separate branch.  You can then make a separate pull request with this change.

You should rollback the change in this PR/branch and then use git push -f to force push.

However, I don't really see the need for this tooltip change.  It adds complexity and isn't very useful (the vast majority of page views are on recent matches).

Aehrraid:
Thanks for the advices concerning the rollback and branches.

I will open a new issue for the tooltips discussion.

PR #553, opened 2015-07-21
Consistent Tags on both blog and blog posts

vramana:


howardc93:
This just turns the current span.label links into regular links?

I'm fine with it, @albertcui should review.

Issue #539, opened 2015-07-18
cal-heatmap wide on mobile


howardchung:
- [x] heatmaps (on match pages and two small ones on player pages) should be scaled to 100% viewport, and heatmap.js scales values by the current width of the image.
- [x] building map should scale to viewport.  
- [x] Building icons should also scale down to match
- [x] rating graph--scrollable and zoomable sometimes makes the page freeze?
- [x] word clouds--locked to 1000px, should scale to viewport (maximum font size may need to be scaled as well)
- [x] player/match pages--nav tabs should collapse like the navbar does
- [x] curly-cue image on faq/blog pages needs to scale to viewport
- [x]  Fix responsiveness of front page (purchase table too wide)
- [x] tables should collapse to avoid stretching the viewport.  Here are some solutions:
http://elvery.net/demo/responsive-tables/ (semi-fixed by table-responsive?)
- [ ] cal-heatmap--too wide for narrow viewports

Aehrraid:
>  Fix responsiveness of front page
Do you mean the general landing page for everyone or the start page for logged-in users (/players/123)?

howardc93:
The general landing page for new users.  I think the issue is the purchases table is too wide and stretches the page.  I worked around it by setting the scaling to 50% so all pages start off more zoomed out, but that's temporary.

howardc93:
We should find a generic solution for displaying our wide tables on mobile.  Once we have one the front page isseu should fix itself.

Aehrraid:
Ok. If i request the front page with 'no-cache' I mostly acknowledge about 5-6s waiting time for yasp.co/ html only. Afterwards, public files take approx. 2s to load. images and icons from cdn.dota2.com take  1.5s and yasp.co images/wallpaper take 1-2s.
*This is worst case*

howardc93:
We have a separate issue for optimizing load times client side: #77

howardc93:
Fixed the table-responsive class not working, so the index page is responsive.

However Bootstrap's solution is simply to make the tables horizontally scrollable without stretching the viewport.  It works, but we can probably display it better.

howardchung:
merge into #368

PR #561, opened 2015-07-24
Align player info and records a bit properly

vramana:
Before:
7fap7sd

After:
ynvvedk


howardc93:
Thanks!   Shouldn't the name be h1 to emphasize it?

vramana:
IMHO, h1 is too large. I would like a more compact UI. I am looking at the following issues.

1. I am investigating why tabs don't collapse well when the viewport is small. 
2. Those labels don't suit well with the theme of website. They seem a bit odd.
3. Filters can be a lot more compact. All the filters under synergies have cursor mis-aligned with tags.

howardc93:
1.  Yes, this is one of the issues in #539 .  We'd love to get some help on that (although we still would have to figure out how to compress the tables for mobile for it to really have an effect).

2.  Go ahead and play around with different display options.  I use them in two places, to display rating on a player profile page, and to display "peak" rating on the ratings tab.  I may move the peak rating to be displayed with the current rating, just to consolidate data...

3.  We are using select2 for the multiple select boxes, and they seem to enforce some weird styling.  I manually override some of the styles here: https://github.com/yasp-dota/yasp/blob/master/public/css/yasp.css#L350

Honestly, I would like to drop it for something more lightweight, and avoid having to use hacky CSS to make it look acceptable.

vramana:
Hey, you tossed around the idea od using WebComponents in the UI improvements issue. I like the idea of componentizing things. Have you ever looked at React? Although, it doesn't solve the CSS global namespace leaks completely, but it allows to write much more flexible components in a declarative way. 

Also, it is really easy to adopt i.e, we don't have to change our existing stack. We can try it out along with our existing stack. We build a nice single page application without having to disturb the system too much.  There also some other benefits we can get out of a complete React based application server-side renderning(Not really a benefit because we are currently doing rending but I think we can do better in some cases), load javascript on demand i.e, we currently dowload around 350kb of javascript, I think we can cut that down so much more. Surely, the home page doesn't need all that JavaScript to render.

I'll understand if you are sceptical about this, I don't mind that. I'll focus on UI for now.

howardc93:
I have never used WebComponents, it's something that @albertcui has been playing with.

I have used React, but I haven't found it very useful for YASP since the user rarely interacts with the data.  We just render it once and it doesn't really change in response to user input.

At this point it would be quite a bit of work to refactor everything into a SPA.  You could try it if you wanted in a small section of the site.

You may want to pull the pm2 branch if you are doing more work.  I changed how the ratings are displayed and am using selectize instead of select2 for the query form.

vramana:
I think that View as JSON should be moved to some other place. Do you have any metrics on how many people click that link or visit that URL?

howardc93:
Ideas for where to move it?  It's not that heavily used, but the feature was asked for by a few people so we just added it since it wasnt difficult.

vramana:
I prefer expose it as a REST API than show it on the page. We just need to write some documentation and link in FAQ and describe it the README of the repo. How about that?

howardc93:
It's not really much of an API, just tack on ?json=1 to a player or match page.  I suppose we could just mention it in the README and remove the link.

vramana:
This is how I see a page
screenshot from 2015-07-24 17 09 08

Almost very little content is being shown. I have to scroll down like an entire page to get to the actual content. I think both the notifications and header takes too much space. 

The Source 2 notification is important but once I have been notified I don't need to see it again and again. And there is now way to dismiss it. How about adding it as tooltip to the skipped status in the match list? I think that would be more useful. 

All though I view the page, I see "This user is tracked! Matches they play are automatically parsed.". It annoys me a little because it should tell "Your matches are being tracked" or something like that. Instead of it big notification at the top, how about showing a small icon with tooltip which mentions whether user is tracked or not. Something like an eye icon. White color would represent the user is tracked, a crossed out dark eye would represent a untracked user profile.

vramana:
Also there is ton of whitespace and large fonts. There is too much padding on above an below the navigation links. 

I nuked a ton of whitespace, removed notifications, reduced fonts size. This is what the page looks like now.
screenshot from 2015-07-24 18 10 06

This is has a lot of rough edges but what do you think? Which one looks better?  On a larger screen I think we should push the adds towards the sides. That would be even better but I am a little afraid that it'll make layout a little bit complex. May be not I don't know. But I think it's worth giving a shot.

howardc93:
Submit a new PR for those changes and I'll take a look.  @albertcui should weigh in, he's the UI guy.

vramana:
Okay.

howardc93:
You are right about the banners though.

The Source 2 banner is actually a "general site notification" banner and is not a permanent part of the site.  We set a key in Redis when we want to display a sitewide message.

I will replace the tracked banner with an icon.

The full history is probably important enough to remain a banner (it also is not as "permanent") since it should disappear once we are caught up.

vramana:
Hey can you tell me how to setup the database server locally?? I really want to look at the changes I make locally. But I have been burnt to crisp trying to setup redis and postgresql in the past. So, I am little afraid to do it own my without any instructions.

howardc93:
We develop on Ubuntu.  Then it should be as simple as `sudo apt-get install mongodb redis-server`

vramana:
Even I am on ubuntu.  My console is filled with these message when I do `npm start` 
```
19:34:14 parseManager.1 |  2015-07-24T14:04:14.209Z - info: retrying: http://localhost:5100/?key=shared_secret_with_retriever
19:34:14 parseManager.1 |  2015-07-24T14:04:14.210Z - info: getData: http://localhost:5100/?key=shared_secret_with_retriever
19:34:14 worker.1       |  2015-07-24T14:04:14.425Z - info: retrying: http://localhost:5100/?key=shared_secret_with_retriever
19:34:14 worker.1       |  2015-07-24T14:04:14.426Z - info: getData: http://localhost:5100/?key=shared_secret_with_retriever
19:34:14 scanner.1      |  2015-07-24T14:04:14.489Z - info: retrying: http://api.steampowered.com/IDOTA2Match_570/GetMatchHistory/V001/?key=
19:34:14 scanner.1      |  2015-07-24T14:04:14.489Z - info: getData: http://api.steampowered.com/IDOTA2Match_570/GetMatchHistory/V001/?key=
```
Can you explain what to do? I am missing some config but i am not able to figure it out from README.

howardc93:
The localhost:5100 error is due to not having a working retriever running.  The retriever requires a Steam username and password.

The other error is due to the scanner not having a Steam API key.

If you are just testing UI changes you don't need either, but you will need some sample data.

You can either run just the webserver with `nf -j Procfile.dev start web`, or add the appropriate env variables in an .env file.

vramana:
Where and How should I add the sample data? I want to work on Profile pages. Is there sample data for that?

vramana:
What does this error mean ?
```
20:13:56 web.1 |  { [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
20:13:56 web.1 |  js-bson: Failed to load c++ bson extension, using pure JS version
```

howardc93:
https://github.com/yasp-dota/testfiles/blob/master/dota.zip contains a sample db dump that you can import.  It has I think about 6000 matches and 28000 players.

I've seen that a few times, but never figured out what caused it, probably lack of build tools on your dev machine.  It doesn't seem to affect operation.

vramana:
I have downloaded the dump. How should I connect it to the server?

vramana:
Apparently, the bson error in my case is due to using python3 as default in which case it fails silently during the installation.

howardc93:
Unzip it and then use mongorestore to import it.  You can read the mongodb docs for detailed usage.

vramana:
After searching for a while, I have tried following this http://www.thegeekstuff.com/2013/09/mongodump-mongorestore/ But no luck. `service mongod start`. Just gives me `mongod: unrecognised service` 

I am sorry for bieng a total noob. Can you point to some basic resources? Just to get up and running.

howardc93:
You could try the mongodb docs.   The service might also be named mongodb, not mongod.

vramana:
The official docs are a bit cryptic. I am reading it hopefully I'll have some success by tomorrow. 

I am trying to play with mobile version. I have put a `hidden-xs` class on a few because it felt like I am cramming too much. This felt a bit clean to me. On the desktop we will have the full description. 

saefcawe

vramana:
@howardc93 Will you merge this or hold off until @albertcui says yes ?

howardc93:
I want @albertcui to take a look first.

albertcui:
This looks great to me. Thanks for the contribution!

vramana:
How about merging totals tab into records tab ?? Totals are calculated in records anyway. It reduce overall clutter IHMO.

howardc93:
They're a little different.  The current "Totals" tab is for multidimensional data (killstreaks, multikills, and wards).

The Records totals are single-value totals.

vramana:
But Totals is kind of Confusing name. Because all Totals are not present only data about Streaks and Obs & Sentries are available. So, its kind of odd that its named Totals. I think Obs & Sentries should be moved to a tab named Vision and the Totals should be renamed to Streaks.That will be a lot more clear than simply saying Totals which doesn't give much info of what we are talking about. 

We don't show maximum no. of double kills or triple kills that happened in a game. But can't that be treated as a "record" ?

howardc93:
I guess we can do that.  Both of those pages will be pretty short, though.

We can treat them as a record, but we have to individually define them.  In `aggregator.js` we have declare the category and then give the path to it.

vramana:
In the current layout everything collapses to a single-column. I have made a few tweaks. Can you give some feedback ?? 


blhf7kx
2sgnwzw
5pa27m0
q9dyo0y
6qkptdo

howardc93:
Depends on what you did.  I think the Bootstrap way is to stack the columns
when the viewport shrinks below the minimum size defined in the column,
e.g. col-md-4.  Setting col-sm-4 or col-xs-4 lets the columns shrink
further before they stack.

vramana:
Yeah I understand that. I have attached a few images. It seems you have replied from your mail directly so, may be you haven't looked at them.

howardc93:
Ah, I didn't see the commit.  Looks good to me, @albertcui will merge when he signs off.

vramana:
I have tweaked things a bit more especially the where cheese appears. 

istthtj
fht9cgg
nf3fpxy
agp4mvn
rrijq7r

howardc93:
I don't really like hiding data on narrow viewports.  Can we figure out a way to display it acceptably?

vramana:
We have decrease the size of text to accommodate things. Probably we can accomdate KDA. Puttting Abandoned will be too much. Layout is will look really wierd. 

On large viewports, progress bar looks nice but as you can see from the last image, the progress bar doesn't look nice at all. On a regular phone size it will shrink even more. Let me try something else.

albertcui:
Aologies that this hasn't been merged in yet. I've been really busy this summer :(

vramana:
@albertcui It's fine. I am not completely okay with current state of PR. I am still looking to improve a few things here and there. I would be happy if you can give some feedback.

albertcui:
I am fine with restricting information. KDA is not important information, esp if you're on a mobile device.

I love the Cheese change. It looks great. I could never get any of the player-header stuff to line up. I'm very impressed :D

albertcui:
I'm wondering if it's a little icon overload though. The rating/mmr stuff doesn't really need to be an icon I think, and the fact that we need to use a tooltip to explain what the icon means kind of supports that. If we can remove the tooltip it makes it a little cleaner, too.

vramana:
> I'm wondering if it's a little icon overload though. The rating/mmr stuff doesn't really need to be an icon I think, and the fact that we need to use a tooltip to explain what the icon means kind of supports that. If we can remove the tooltip it makes it a little cleaner, too.

Even I think so. I think even peak solo/party mmr, are not so important there.

What do you think about the progress bar? It really looks ood to me on smaller viewports especially. It doesn't align well which I think makes it stands a bit odd among everything else. 

Is there any reason to use uppercase everywhere all the headers & player name?? Can you explain the rationale behind it?

I want to decrease the over all font size over the entire page by 10% or so.

howardc93:
We could display MMR only on the rating tab.  I figured it was of interest to most people so I put it on every page.

I think the progress bar looks fine but I'm open to different ways of displaying winrate.

@albertcui did the uppercase thing.

vramana:
I tweaked the filters a bit. Now they don't occupy as mush space as they used to.

Before:
slonw0e

After:
cz7oc27

howardc93:
The problem with no labels is that the multiple select fields become
ambiguous when populated

Brian-McG:
> The problem with no labels is that the multiple select fields become
> ambiguous when populated

Agreed, the filter can be collapsed so the additional white-space is less of an issue.

vramana:
I couldn't come up with a better solution than this. I tried a changing the layout completely but it didn't come out well.  

@Brian-McG How would you collapse the filter?  I think only for synergies the labels are needed in other forms fields the placeholder themselves explain the form well. We dont need to repeat that.

@albertcui Can you give some advice?

vramana:
Oh sorry. There was another element with the same id so, I changed it. I'll change that back and change the other elements id.

howardchung:
The form changes are separate and should be rolled back in this PR.

I made some changes myself to forms which address some of the issues (using selectize for all fields, fixing the id conflict)

This branch should only contain the changes to alignment and columns.

vramana:
Okay.

albertcui:
Merged this without the form changes. Thanks a lot! Sorry again about the wait.

PR #573, opened 2015-07-30
reduced homepage weight

howardchung:
Reduced bandwidth requirement for homepage by removing sections/background images.

@albertcui should review, I may have been a little heavy-handed.

Saves:
* 400KB, our mission section
* 430KB, positions map
* 400KB, details background image


howardchung:
I am going to merge this soon unless @albertcui pushes back.

I don't know if you're proceeding with the retheme (and how much it affects the home page) but you can fix it later.

It's important to have a fast-loading index page and not make new visitors have to sit through a 2.5MB load.

PR #576, opened 2015-08-02
fixes #566

### LINKED ISSUE #566: move processParse to parse nodes
@nickhh has mentioned this before, we can discuss this issue here.

Currently the parse nodes can run completely statelessly (without any environment variables required).  They act solely on the data within replays and transform it into newline-delimited JSON that is sent back to the client, where it is post-processed into a single parsed_data JSON blob that we store in the database.

Moving the post-processing to the parse node is an idea suggested by Nick.

Implementation:
We would have the parse server send back an object with type "parsed_data" as the final line.  Prior to this it could send back "progress" in order to keep the client updated.

On the client side, receiving an event of type "parsed_data" allows us to set error to null (the parse succeeded).

Most of the runParse function would be moved over to parser.js, with the only types left being "progress", "error", and "parsed_data".

Pros:
* Less bandwidth usage.  The parsed_data object is much more compact.
* More distributed CPU workload.  The parsing "farm" contains more CPU power than YASP core, and probably can afford to spare more time crunching the event buffer into JSON.

Cons:
* The parse nodes must have a copy of the constants in order to do post-processing.  They are required to map combat log names "npc_dota_hero..." to a hero_id, which can be used to match the combat log name to a player slot.
* Not "streaming".  The client must wait for the entire replay to be processed before getting data back from the server (although we can somewhat mitigate this by emitting "progress" events based on download progress)
* A bit more duplication.  We would need to make use of the ndjson module on both ends to deal with the streaming JSON and also set up the event handler structure that picks up events and processes them.

howardc93:
Loses the ability to stream progress, but less CPU intensive on yasp-core


howardc93:
@albertcui and @nickhh should decide if this tradeoff is worth it.

Others can feel free to weigh in as well.

nickhh:
I think the progress bar is nice (as long as it's not buggy). Is the
speedup worth it? Do we have statistics on how many games are requested
like this?

howardc93:
It's not really a speedup, just less CPU/RAM usage on yasp-core.
I think it's a few hundred requests a day.  I could get  more precise numbers by running a db query.

nickhh:
So it's like less than 1% of parses per day? I feel like it's not worth it,
but that's just my opinion. Which do you prefer, slash why?

howardc93:
I'm not sure you understand.  We CURRENTLY have the ability to stream progress updates to the user since the parser emits progress events.  

With this PR, by moving the logic that combines the JSON events into a single object we lose the ability to output progress, so the bar would appear to jump from 0% to 100%.

I cannot say that one is definitively better than the other.  It's a question of tradeoffs (balanced CPU usage/less bandwidth vs. no progress updates/maintaining additional state across multiple nodes)

nickhh:
I know. I meant to say that since these requests are such a small
percentage of the cpu's load that it's not worth optimizing it's
performance at the cost of cool features like status updates.

howardc93:
This affects all parses not just requests.  The load is pretty significant.

nickhh:
Ohhh I see. So you're decreasing load for all parses, and loosing a
progress bar for 1% of them. Then yea, I think if you think the decrease in
load is worth it, I'm all for the change :) How did you decrease the load,
by the way?

howardc93:
Instead of combining the JSON on yasp-core it is combined at the parse node
and only the resulting single object is sent over to yasp-core.

PR #580, opened 2015-08-09
Steamv1

howardc93:
Refactored retriever to use Steamv1.

node-dota2 still requires 0.6.8, and will require some significant refactoring to make it compatible with v1.0.0.


howardc93:
NVM, this is not ready for merge yet.  Since the retriever creates the Steam client object and passes it to node-dota2, node-dota2 receives a Steam object of v1.0.0.  Therefore this can't be merged until node-dota2 is refactored to support steamv1.

PR #590, opened 2015-08-20
Steamv1

howardchung:
https://github.com/RJacksonm1/node-dota2/pull/90

The above pull request is supposed to make node-dota2 compatible with node-steam v1.1.0.

If it works for matches/players (needs to be tested), we could:
* If node-dota2 maintainers accept it, use official node-dota2 module
* If they don't accept it, pull it into yasp-dota/node-dota2

Then this branch can be merged (with update to package.json).


howardchung:
Okay, this should now work, but I am using some hacks to make node-dota2 work.  It's mergeable, but the library will probably be cleaned up more later.

howardchung:
Tested to work with matches and players under node-steam v1.1.0.

Retriever now restarts itself if any account is logged out or a steam error occurs.  Don't know how this scales.

Otherwise, this branch should now be mergeable.

package.json still points to yasp-dota/node-dota2#steamv1.  I think this is okay until node-dota2 merges the changes in, then we can swap over to official node-dota2 in package.json and delete our fork.

PR #593, opened 2015-08-22
Enable full-game position heatmap

howardchung:
Not a difficult feature to light-up, but we need to carefully consider the storage cost.

The match I examined went from 180k to 320k.  Compression helps, but it's still a heavy cost (and I don't think the position data compresses very well since it's basically all numbers)


howardchung:
We could attempt to reduce size by limiting resolution.

Divide x/y values by 2 and round down (using ~~), then rescale it up by 2.  Essentially, we force all values to be a multiple of 2, reducing the size of our x/y hash (more overlaps).

howardchung:
latest attempt: 156->192kb

parser.js
```diff
@@ -533,14 +533,18 @@ function runParse(data, cb) {
...
                 }
-                //add to positions
-                //not currently storing pos data
-                // if (e.x && e.y) {
-                //     e.type = "pos";
-                //     e.key = [e.x, e.y];
-                //     e.posData = true;
-                //     //populate(e);
-                // }
+                e.interval = false;
```
howardchung: unnecessary

Issue #511, opened 2015-07-09
enable toggling between timeseries and per-match MMR graphs


howardchung:
feature requested on reddit

howardc93:
resources for c3:
http://c3js.org/examples.html

PR #632, opened 2015-09-12
Actions

### LINKED ISSUE #454: tool-tip clarity/consistency
update tool-tips across the site for clarity and consistent style

TODO:
- [x] externalize tooltip strings
- [x] tooltips for categories in player pages (histograms, records, trends, compare)
- [x] tooltip for "status" column in player match tables

### LINKED ISSUE #591: sentry and dusts are not tracked as sets
hi,
Currently the tooltip for sentries and dust bought(http://yasp.co/matches/1730394899/purchases) ends with "(x2)" which, would imply that to get the real amount of sentries/dusts you would you have to multiple the number by 2. This is not the case, so I suggest just removing the (x2).

### LINKED ISSUE #596: trends/over time statistics
I suppose it would not be too hard to supply a "rolling" average of all the stats we already keep track of (same set as histograms, records, compare).

As we iterate through a player's matches, for each stat we can take the current sum and divide by the current N, and push the number into an array.  The resulting array can then form a trend/over-time graph of the player's performance in that stat.

The only potential sticking point is how this might affect cached players when we add matches out of order (though request form or filling in through full history).  

Potential solutions:
* Flush player caches when we add a match with that player.  Makes the cache not that useful if we clear it every time someone plays a match.
* Store full parsed data for matches in the player cache.  Then we can sort the matches at page view time.  Memory intensive (unless we move our caches off Redis and onto disk).  Maybe we should use a separate collection in mongo for this?

howardchung:
fixes #454 
fixes #596 
fixes #591 

Actions tab could use some cleaning up (shorter strings, don't display every action type).

There are 29 action types and not all of them are interesting.

TODO
- [x] add "new" tags to actions and trends tabs
- [x] lightup mmstats
- [x] disable mongocache
- [x] make parser parallelism constant config value?
- [x] split streaks and wardmap into separate tabs
- [x] fix weird roundoff error with tp scroll purchases on trends (1.400000...1?)
- [x] ~~display formatted time instead of seconds for duration/firstblood~~ (meh, maybe later)
- [x] write release notes

POST-DEPLOY:
- [ ] flush redis caches (new aggData): `redis-cli keys "player:*" | xargs redis-cli del`
- [ ] make sure new code deployed to all parse workers
- [ ] monitor for process leak on parsers


howardchung:
Actions tab should be okay, @albertcui should review.

Trends tab could use a little improvement.  I'm currently just storing data points in an array.  If we want to add data to the x-axis such as the date of the point, then we have to store an array of objects instead.

We'd also have to figure out if we want to display the graph as timeseries or by index.

This is similar to the issue that @lemongod is working on: #511 , where he's implementing a toggle.

PR #635, opened 2015-09-13
Trends

### LINKED ISSUE #596: trends/over time statistics

howardchung:
fixes #596 

Can use some styling before deploy.


howardchung:
merged into #632

PR #649, opened 2015-09-15
parserleak

### LINKED ISSUE #543: memory leak on parse server
I think parser.js has a memory leak where Java processes aren't being cleaned up.

It should kill the children when it encounters an error while parsing, but apparently this isn't picking up 100% of cases.

Also, the entire server should not be wrapped in a domain so that the parser can crash and restart.

howardchung:
fixes #543 

Not tested yet.


howardchung:
I have no idea if it solves the problem, but it parses.  It's hard to test this in development.

PR #656, opened 2015-09-17
Show unchanged levels without arrow

dancek:
Currently the Teamfights page always shows levels like 9->9. Change that to just 9, but show actual changes like 9->10.

### DIFF
--- a/views/match/match_teamfights.jade
+++ b/views/match/match_teamfights.jade
@@ -102,7 +102,7 @@ block match_content
                 if player.buybacks
                  div(style="font-size:30px;") !{"&#128176;"}
               td(style="white-space: nowrap;")
-                if player.level_start != player.level_end
+                if player.level_start !== player.level_end
                   | #{player.level_start}<i class="fa fa-long-arrow-right"></i>#{player.level_end}
                 else
                   | #{player.level_start}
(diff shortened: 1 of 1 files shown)


views/match/match_teamfights.jade
```diff
@@ -101,7 +101,11 @@ block match_content
               td
                 if player.buybacks
                  div(style="font-size:30px;") !{"&#128176;"}
-              td(style="white-space: nowrap;") #{player.level_start}<i class="fa fa-long-arrow-right"></i>#{player.level_end}
+              td(style="white-space: nowrap;")
+                if player.level_start != player.level_end
```
howardchung: use !==

dancek: Consider 9 != '9' where I'd still leave the arrow out since that would render as 9->9.

howardchung: The types should match.  If they don't we're doing something wrong in our code somewhere.

Generally I think it's better practice to use !== and do any necessary type conversion yourself.  Otherwise it's easy to have bugs.

dancek:
I see your point, here's the fix. I was lazy and did this with the online editor. Do you want me to rebase these into one commit?

howardchung:
you can if you want but it's fine to leave it.

Not sure if the pipes are needed but they probably don't hurt either.

@albertcui should review and merge.

dancek:
I'll leave it like this, then.

I do think you need the pipes because Jade expects the first token on a line to be the tag name (even if it's a variable).

albertcui:
Thanks for the contribution!

PR #658, opened 2015-09-19
Java parser code cleanup

mbax:
I rather enjoy using yasp, so I figured I would see what's under the hood. Seems there have been some scary creatures growing there! Figured I'd help clean it a bit first. Explanation of each commit is below. You'll want to look at the individual commits to see what on earth is going on since so many lines have changes due to the first two commits.

1. [Replace tabs with spaces for consistency](https://github.com/yasp-dota/yasp/commit/6117f8741bdff43c0fc11b6c29dbd69f5cd32a46)
    * There was a scattering of tabs and spaces in use. Consistency makes for happier coders.
2. [Consistent spacing between elements](https://github.com/yasp-dota/yasp/commit/ff80fc5a00dd47602bc2091e1ea67dd4c8335515)
    * One could find statements like `if(true) {` `if (true){` `if(true){` and `if (true) {`
    * Consistently applied spaces for readability.
3. [Built with Java 7, utilize Java 7 features](https://github.com/yasp-dota/yasp/commit/3072d86d7728488df97e26b50d77d17371bb387b)
    * Syntactic sugar like Java 5's foreach and Java 7's diamond operator for generics sweeten the codebase.
4. [== on String is not consistent](https://github.com/yasp-dota/yasp/commit/05fba0879be84ebf7a08e9250ffe1717e66a338b)
    * One can't be sure that calling == on a String will result in equality unless you are certain it is being `intern()`'d and that's a gamble best avoided.
5. [Remove unnecessary casting](https://github.com/yasp-dota/yasp/commit/bb1d720f808a79db6521d96ea568b8eea1e27774)
    * Dropped a lot of unnecessary casting, thanks to cfca84e035f0a856471e1a3a6aa0fa84d5d06d25
6. [Reduce redundant code](https://github.com/yasp-dota/yasp/commit/494d9664d9ccbbce34aa19613b99151f4b3023c9)
    * Dropped the `if (true)` statements
    * Merged two foreach blocks that didn't need to be separate
    * Removed unused imports
    * Removed unnecessarily declared variables, such as:
        * Declaring a new reference for the same variable
        * Declaring a variable as an intermediate point in a chain of method calls
7. [Order imports](https://github.com/yasp-dota/yasp/commit/2b0215d126e98c27b1dc691d3e3021ccb7d2a0a6)
    * That was just bugging me

### DIFF
--- a/java_parser/src/main/java/yasp/Entry.java
+++ b/java_parser/src/main/java/yasp/Entry.java
@@ -33,9 +33,11 @@ public class Entry {
     public Integer y;
     public Float stuns;
     public Integer hero_id;
-    public Entry(){
+
+    public Entry() {
     }
-    public Entry(Integer time){
+
+    public Entry(Integer time) {
         this.time = time;
     }
 }
--- a/java_parser/src/main/java/yasp/EventStream.java
+++ b/java_parser/src/main/java/yasp/EventStream.java
@@ -1,14 +1,18 @@
 package yasp;
+
 import com.google.gson.Gson;
+
 import java.util.List;
 
-public class EventStream{
+public class EventStream {
     private Gson g;
     private List<Entry> es;
-    public EventStream(){
+
+    public EventStream() {
         g = new Gson();
     }
-    public void output(Entry e){
-        System.out.print(g.toJson(e)+"\n");
+
+    public void output(Entry e) {
+        System.out.print(g.toJson(e) + "\n");
     }
 }
--- a/java_parser/src/main/java/yasp/Main.java
+++ b/java_parser/src/main/java/yasp/Main.java
@@ -1,35 +1,34 @@
 package yasp;
 
+import com.google.gson.Gson;
 import com.google.protobuf.GeneratedMessage;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import skadistats.clarity.decoder.Util;
 import skadistats.clarity.model.Entity;
+import skadistats.clarity.model.FieldPath;
 import skadistats.clarity.model.s1.GameRulesStateType;
+import skadistats.clarity.processor.entities.Entities;
+import skadistats.clarity.processor.entities.OnEntityEntered;
+import skadistats.clarity.processor.entities.UsesEntities;
 import skadistats.clarity.processor.gameevents.CombatLog;
 import skadistats.clarity.processor.gameevents.OnCombatLogEntry;
-import skadistats.clarity.processor.entities.Entities;
-import skadistats.clarity.processor.entities.UsesEntities;
-import skadistats.clarity.processor.entities.OnEntityEntered;
 import skadistats.clarity.processor.reader.OnMessage;
 import skadistats.clarity.processor.reader.OnTickStart;
 import skadistats.clarity.processor.runner.Context;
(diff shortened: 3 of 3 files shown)


albertcui:
Thanks for the contribution! I'll like @howardchung look it over since he's written most of the parser. I believe most of the issues are because he doesn't develop in an IDE... >.>

mbax:
Most of what I contributed is purely about making it look cleaner, which I can certainly see being a result of [not using an IDE](http://i.imgur.com/RYXLbWg.jpg). 

05fba0879be84ebf7a08e9250ffe1717e66a338b however might make something behave properly and the removal of a double for-loop in 494d9664d9ccbbce34aa19613b99151f4b3023c9 should shave off *at least* a nanosecond from execution!

howardchung:
Yeah, I develop using Cloud9 (c9.io) and unfortunately the Java support is terrible :(

spaces-> tabs looks good!

consistent spacing-> looks good!

Java 7 features: looks good, but make sure it builds.

== on String: Yep, you're right.  It's hard to remember to have to do that in Java since it's not necessary in JS.

Unnecessary casting:  Just gotta make sure it builds.  I wasn't sure if the casts were necessary--was basedd off @spheenik 's matchend example (which I think uses explicit casts).

Redundant code:
Reason for some of the double-assignments were for logging (while developing I might have wanted to print those values).

The reason I was using if (true) was that I was getting build errors about redeclaring the entry variable.  Not sure if that's still a problem.

Order imports: Looks good!

As a warning: As development continues imports may end up out of order again, etc, due to my lack of IDE.  Spaces should remain consistent since I think the tabs are probably left over from the examples we based it off.

howardchung:
By the way, thanks for splitting up your commits.  Made it very easy to review :)

mbax:
> Java 7 features: looks good, but make sure it builds.
I confirmed everything happily built as I was working on it. Plus, IDEs these days will whine quite a lot if building is going to fail.

> The reason I was using if (true) was that I was getting build errors about redeclaring the entry variable. 
That's because you tried to declare the same variable name twice. I renamed all the entry variables there so it was clear which was being touched. An alternative solution was to reuse the variable but that gets confusing when reading. 

> As a warning: As development continues imports may end up out of order again, etc, due to my lack of IDE.
The goal was just making it look a bit cleaner. If it starts to fall back again there's always further cleanup commits. I sometimes end up doing the same on my projects when I forget to check if my code matches my own standards.

> By the way, thanks for splitting up your commits. Made it very easy to review :)
You're quite welcome. I find it easiest to split big changes into individual steps, so the actual changes are easier to see. If this were one commit you'd miss most of the changes because everything would be diff due to the spacing changes.


On a related note, do you folks have any place that you communicate about this project outside Github issues (and the blog, which is a one-way communication)?

howardchung:
There's the Gitter chat, although it hasn't been used in a while.
https://gitter.im/yasp-dota/yasp?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge

There's also Twitter, I guess.
https://twitter.com/yasp_dota

howardchung:
signoff, @albertcui merge at will

albertcui:
Thanks for the contribution!

Issue #676, opened 2015-09-25
distributed parse client


howardchung:
Currently parse workers are completely independent and stateless, acting as servers rather than clients.

They just respond to http GETs that give them a URL and return the JSON parsed data for that match.

This means a couple of weird things:
* yasp-core has no idea how much processing power a given worker has, and just blindly assumes some configurable number (currently 8 parallel parses)
* We can't adjust the number of parallel jobs on the fly without stopping, updating our config, and restarting.  There is an issue in kue for this (https://github.com/Automattic/kue/issues/447) but it hasn't been implemented.

Reversing the parse flow means that we open up an endpoint on yasp-core that RECEIVES parsed data uploaded as POST.

Parse workers get a URL and key (to prevent abuse) set in config and GET from the endpoint, requesting work.  (This may be annoying, I don't think kue supports the ability to "pop" a single job off the queue) yasp-core sends back the URL of the replay file. That job should be marked "active" in kue so we don't reassign it to another parse worker.  The parse worker uses that URL, downloads/parses the file, and POSTS the resulting JSON back to the same endpoint.

yasp-core then marks that job as complete when it receives a POST that contains the data for that match.

We might need to add a timeout for the job if the parse worker never finishes it.

Advantage:
* We can add/remove parsers dynamically without touching yasp-core.
* Potentially could open it up publicly so anyone with a server can contribute parsing power.  We would likely to need to implement some kind of double-checking to prevent abuse, though.  If we made a leaderboard and implemented some kind of scoring system, we might get a lot of interest.
  * Making sure everyone is running the correct, latest version of the parser may be annoying

Disadvantage:
* We have to set config values on each node we want to add.
* How to handle case when queue is drained?  Do we send back a null value for the replay URL, and client tries again in some timeout interval?

Medowar:
You mention that yasp has the number of concurrent Processes hardcoded somewhere. Can you please tell me, where that value is getting defined?

howardchung:
Currently here:
https://github.com/yasp-dota/yasp/blob/master/config.js#L39

It's configurable with a .env file setting though.

Medowar:
Thanks. I did turn down the number from 8 to 4 for my Parsing Server for yasp.co

howardchung:
Actually that may not do much with the current code.  We still send 8 simultaneous requests from the server.  #582 should fix this and I'm hoping to get that out by the end of the month.

howardchung:
Fixed in #685

Issue #667, opened 2015-09-23
update example match


howardchung:
The example featured on home page is quite old.

howardchung:
maybe wait after sql migration since the format is incompatible and you'll have to update the json anyway

howardchung:
fixed

PR #685, opened 2015-09-29
Sql

### LINKED ISSUE #582: PostgreSQL migration/engine upgrades
A relational DB is probably a better option for Dota match data.

Something we could play with/try out if somebody wants to take the time:

Checklist moved to PR

### LINKED ISSUE #625: why disallow /players in robots.txt?
When I go to friends' places to show them my profile, I have to log in via steam instead of just googling my name + yasp :(

### LINKED ISSUE #679: split histograms/trends into subpages
While @Aehrraid 's work has somewhat improved this by not rendering all the charts at once, the fact remains that the page still loads the data for all of them and this can take quite a while to do.  I think we need to split up each histogram/trend into its own subpage.

While this means we can't do the toggle-comparing, I'd really like those pages to load faster (and it's also kind of wasteful to send all the data when the user probably isn't interested in all of the histograms/trends).

### LINKED ISSUE #681: missing cluster 121 (us east)
Looks like data files are missing this one.  We can manually patch it in.

### LINKED ISSUE #686: remove 0% label from histogram if n=0
http://yasp.co/players/88367253/histograms

### LINKED ISSUE #719: vision tab
Should we consider moving all vision related stats to their own tab?

Right now things are bit scattered:

* Dewards are under "Farm" tab since that's where the rest of the kill data is
* ward purchases are under "purchases" since that's where the rest of the purchase data is
* ward positions are on the heatmap since that's where the heatmap is

We could possibly get rid of the position tab, move ward positions to vision tab and move laning heatmap to performances tab?

howardchung:
fixes #582 
fixes #681 
fixes #625 
fixes #686 
fixes #679 
fixes #719 

DEPLOYMENT SCHEDULED: 2015-11-06

DEVELOPMENT
----
- [x] Figure out a schema (parsed_data, player_match and player_rating should probably be separate tables)
- [x] write migration script and test on devboxes
- [x] pass single reference to db/redis
- [x] update views to not ref parsedPlayer, parsed_data
- [x] rename parsed_data.players.gold, lh, xp -> (gold_t, lh_t, xp_t), views
- [x] rename parsed_data.players.kills -> killed, views
- [x] rename last_visited -> last_login, views
- [x] change player rating fields from camelcase to snake case (soloCompetitiveRank -> solo_competitive_rank)
- [x] queries.js insertmatch, when inserting player_match select by match_id, player_slot to ensure uniqueness (account_id doesn't work since anonymous)
- [x] aggregator.js, update to not ref parsedPlayer
- [x] filter.js, update to not ref parsedPlayer
- [x] filter, aggregator should accept player_matches, fellows as first arguments
- [x] status.js, update queries
- [x] mmr.js, update insert query
- [x] buildSets.js, update queries
- [x] getReplayUrl.js, update queries
- [x] routes/donate.js, update queries
- [x] routes/matches.js, update queries
- [x] routes/players.js, select from player_matches join with matches then make separate query for played_with/played_against
- [x] tasks/fullHistory.js (low priority)
- [x] ag -i "mongo"
- [x] ag -i "parsed_data"
- [x] ag -i "parsedPlayer"
- [x] test foreign key (dropping a match should drop all associated player_matches)
- [x] figure out why players show up as bot (account_id=null) when the player record doesn't exist (should use account_id of player_match)
- [x] cache of "all" limited to 250 player_matches, only 25 results due to conflicting match ids
- [x] fix tests
- [x] bigint or integer for match_id and account_id? (has to be bigint since integer only goes to 2b)
- [x] swap kue for bull?
- [x] implement distributed parse clients #676 
- [x] UPSERT not supported until psql 9.5, currently queries use insert, if err, update
- [x] enable stop/resume in migration script
- [x] update README
- [ ] race condition in cheese donation code (@albertcui)
- [ ] merge #715 
- [ ] #667 or else the homepage will break? (@albertcui)

TEST
----
- [ ] make sure donation works (try in sandbox mode)
- [ ] make sure signin works (manual test)
- [ ] make sure requesting an existing match works
- [ ] make sure requesting a new match works
- [x] make sure sequential parsing works (enable pro parsing)
- [ ] make sure cache aggregation works
- [x] make sure old matches don't crash
- [x] make sure fullhistory works
- [ ] make sure new deploy process works

UPGRADE
----
Upgrade to latest node/npm (4.2.0, 3)
`sudo npm install -g npm && sudo n latest`
- [ ] yasp-core
- [x] parsers (contact admins for boxes where we don't have sudo access)
- [x] devboxes
- [x] engine in readme
- [x] travis.yml
- [x] fix problem with build: contextify->nan->microtime->sifter->selectize, contextify->d3 3.5.0->c3
- [ ] redis->3.0.4 `sudo add-apt-repository ppa:chris-lea/redis-server && sudo apt-get update && sudo apt-get install redis-server`

DEPLOYMENT
----
- [x] take a snapshot
- [ ] swap to hm-4 instance
- [ ] Install postgresql on production/dev
- [ ] write release notes/blogpost
- [ ] disable caches
- [ ] update yasp-core (npm install, etc.)
- [ ] dump caches: `redis-cli keys "player:*" | xargs redis-cli del`
- [ ] clear kue keys
- [x] check for null account_id/match_id in production
- [ ] update parsers
- [ ] set `WORK_URL`, `RETRIEVER_SECRET` on parse nodes
- [ ] Migrate existing players `node migrations/sqlMigrate.js players`
- [ ] migrate existing matches `node migrations/sqlMigrate.js matches`


Mysrt:
FYI im not sure what your use-case is for upserting, but one way around upserting is to first do a delete, then an insert.  This removes the possibility of the race condition, if you truly must do an update/insert you can/should lock the table you are updating and then do the update/insert.

howardchung:
Delete won't work since we might get rid of parsed data if reinserting from
API.

Mysrt:
do you update the parsed data in batches?  You can always just add a unique index if one isn't there already and just insert.  If it's already there catch the error.  I looked through it briefly, obviously im not very familar with the codebase, but it didn't seem like an upsert was necessary.

howardchung:
Parsed data goes into the same table (diff columns) as API data.

Mysrt:
hmm so your could normalize this data to fix the problem.  I can see the use-case for upsert in the meantime.  Since they are updating different columns i don't think you will have problems with race conditions though if you choose not to normalize the data.

howardchung:
Some fields are mutable on updates.  For example if a player was anonymous
at first but then enables third party data, we want to be update the
relevant player_matches with their account id.

Similarly players logging in should update their row which could exist
already.

I am doing insert, if err, update right now.  But once we have native
upsert it is probably better to use that

gortiz:
>
> i don't think you will have problems with race conditions though if you
> choose not to normalize the data.


No, you can't. On postgresql, every row update is in fact a row insert on
disk, so if you have 2 concurrent row modificatios, you have 2 inserts on
disk and there will be a race condition.

The solution used by Howard is ok if the code is inside a try-catch-loop to
avoid race conditions. As they are very infrequent, it shouldnt be a
performance problem

2015-10-02 22:39 GMT+02:00 Howard Chung <notifications@github.com>:

howardchung:
I think this solution is actually free of race conditions.  In the case of conflicting inserts they will simply both try to update and then "last writer wins".

The only problem with it is that it can cause insert exceptions to be swallowed.  The update will probably no-op (if we were inserting a new row), which means the insert doesn't actually happen but we don't report an error.

gortiz:
I am not sure if it is free of race conditions. Can you show the code snipet? The "last writer wins" approach will throw an exception and in some cases it can lose data. I am assuming you are doing all modifications on a transaction with a safe isolation (TRANSACTION_READ_COMMITTED or TRANSACTION_REPEATABLE_READ) mode. The following code is incorrect because an exception can be thrown while trying to commit if two threads are modifying the same row at the same time:

'''
    doModify(row)
    startTransaction()
    try {
        insert(row)
    } catch (whatever ex) {
        update(row)
    }
    commitTransaction()
'''

If you ignore that commit exception, your change will not be preserved.

howardchung:
Example:
https://github.com/yasp-dota/yasp/blob/sql/queries.js#L115

The insertMatchTable, insertPlayerMatchTable, and insertPlayer all basically do similar things.

I don't have any explicit transactions in the JS.  I think the default is to commit each transaction as a single command.  So the insert gets committed (if successful), then the update.  If the insert fails, the error comes back to JS and we issue an update.

gortiz:
I am not used to JS syntax or APIs (go! go! Java! ^_^) but if you are not
using transactions, you can be sure there will be race conditions. If you
have a thread A that reads a row and modifies column *a* and a thread B
that reads the same row and modifies column *b*, then, if you are not using
transactions, the execution could be something link that:

thread A: reads row
>
thread B: reads row
>
thread B: modifies b in memory
>
thread A: modifies a in memory
> thread A: stores a
>
thread B: stores b... without the changes done by A
>

So you think everything is fine but you lost the change made by A. In
general it is a problem. It could happen that in this specific scenario
that is fine, but it is not a good practice. This is one of the things you
didnt think about when you use a database like MongoDB because you run your
code and everything looks fine, but then you lose your data and have a big
problem. I am sure JS has an API that let you use transactions. It is
vital.

The same code, with transactions (and a safe isolation level), will throw
an exception when thread B tries to store its not-updated data. Then you
can catch it and try again.

2015-10-05 18:47 GMT+02:00 Howard Chung <notifications@github.com>:

howardchung:
Yes, it does have a transaction API.  (http://knexjs.org/, basically just a library to make query generation a little easier, I felt that a full ORM was unnecessary) 

However, we actually never do any read/modify/update workloads for YASP.  We always get our data from an external "source of truth".

Three cases where we update db data:
* We get match data from API and can insert/update it directly.
* We get match data from the parser and can insert/update it directly.
* We get player data from Steam OAuth and can insert/update it directly.

You're right about it not being a good practice, though, and actually I just thought of possibly the only place in our code where it matters:
https://github.com/yasp-dota/yasp/blob/sql/routes/donate.js#L94

We just take the amount of cheese in req.user, add the donation amount, and update.  There's a race condition there if the same player attempts to donate twice in a short period of time.  (I think this actually might affect us already with MongoDB--it's not a database specific problem).

gortiz:
I am sure you had the same problem with MongoDB. In the last MongoDB World at NY there was a talk about race conditions and concurrency problems and how to fix/avoid them... it is insane the things you have to do with MongoDB to execute isolated updates!

But I dont understand how can you update something without reading it first. I mean. If the row was not inserted before (= if there is no other row with the given pk or whatever you are using to define the "equality" of rows) everything is fine, you insert and that is all. But if the row was there before, then I guess you need to read the stored data and merge it with the changes you get from the external source. There are two cases when you dont need to do that:

* When you do not modify the same row twice (but this is not the case, otherwise you wouldn't need the update phase).
* When the row is always calculated from the same external source, meaning that if there are more than 1 concurrent write to the same row, they will write the same.
* When you dont care about losing some very recently modification to the row. 

I mean: lets thing a row is modified concurrently by a thread that is reading form the API and, at the same time, but a thread that is reading the parsed data. Lets say the row has 3 columns: (pk, A and B). If the API thread tries to insert (pk1, A1, null) and the parser thread tries to insert (pk1, null, B1). If you do not read and merge before insert/upsert and the threads run at the same time, the stored row will be always (pk1, A1, null) or (pk1, null, B1) meaning that you will lose the work done by the API thread or the one done by the parser thread. I can imagine that usually the parser thread will write several seconds after the API thread... but it is not safe to assume that there will be always the same threads, they will be always executed in the same order, etc. I am almost sure it is safe to suppose that right now but who knows about the future! Ok, this is not a critical application like a bank or something like that. Overengineering is not a good thing and I dont think you must change that with top priority, but it is good to have in mind that the program could fail there if, in the future, the context change.

howardchung:
We don't need to read the existing row because the latest data will contain all the fields we care about inserting/updating.

According to Postgres documentation: http://www.postgresql.org/docs/9.4/static/sql-update.html

UPDATE only affects the specified columns and leaves other values intact.

When we write API data, whatever we got from the API should be whatever we write.  We don't care about anything that may or may not exist in the DB.  So we can blindly upsert safely without data loss.

This isn't strictly the case for parsed data (it won't contain everything the API gives).  However, due to the fact that omitted columns are preserved during UPDATE, the example you mentioned should always result in (pk1, A1, B1) no matter what order the operations happen in.  The key is that we don't have to UPDATE with a full row with NULLs for missing values.

(I am not sure if this will change with native UPSERT in postgres 9.5).

In practice the parser will update 5-10 minutes after API.  In my current implementation INSERT is only done on API data.  Parser data and skill data (which comes from another API) is UPDATE only, meaning that it will no-op if for some reason we don't have that match in the database.

gortiz:
Ok, if you are only updating the modified columns and the columns modified
by different threads are disjunt, then it should work. It would be a
problem if at least one column was modified by both threads, as it is usual
on ORMs.

2015-10-06 12:13 GMT+02:00 Howard Chung <notifications@github.com>:

howardchung:
I think the columns will either be disjoint, or constant, so the insert, err, update loop should be sufficiently safe for our purposes.

I think with native UPSERT, the entire operation is atomic, so even if we were modifying the same columns, it would simply be a "last writer wins" case with no data loss (assuming we don't need to read as part of the UPSERT).

howardchung:
@gortiz and @ahockersten :

Do you guys think it makes more sense to make parsed player_matches a separate table (joined with player_matches on (match_id, player_slot), or keep the current model of a single table with NULLs for unparsed matches?

Some current numbers for reference:

76 million matches in DB (~760 million player_matches)
~2.4 million parsed matches

At some point we might import every match ever played--this would be roughly 650 million matches (~6.5 billion player_matches).

gortiz:
I am checking your DSL right now. What is what you call "parsed" from
player_matches? If you are talking about all fields after *additional_units*
then I definitely think you should externalyze that information to another
table if only 1 per 30 games are parsed.

On the other hand... why are you using the type json to store JSON
documents instead of jsonb, which is usually better? I would recommend you
to store on tables some of the information you store on json columns. I am
not sure, but it seems that almost all parsed-related-columns can be easily
stored in specific tables. You can improve your storage eficiency a lot if
you do that. Even more, Yasp web could be faster without using JS caches
because for example when you load the "combat" tab, you could fetch all
combat related with one single and very fast access to the db.

2015-10-20 8:53 GMT+02:00 Howard Chung <notifications@github.com>:

ahockersten:
@howardchung I don't have enough experience with database design to offer a solid recommendation, unfortunately. However, since "decent database design" is on my list of things I need to learn, I did some reading on database normalization, and some thinking. Here's what I came up with, but take it with heaps of salt, because as I said this is what I came up with after an hour of googling and reading, not something based on years of experience working with databases and having a good gut feeling of what's expensive and not.

Also, expect to read for a bit, there will be some thinking aloud, and it got a lot more thorough than I expected.

I will call the two models the two suggestions the "non-separated" and the "separated" cases, just to have something to refer to

The two designs each have different possible costs associated with them:
- non-separated case will (currently) store 32 NULLs for each player_match
- the separated case will require JOINs in a bunch of places where it would be avoided otherwise

## Data storage costs

Let's do some calculations with data I will make up along the way, hopefully without too outlandish assumptions (please question these! I'm pulling numbers out of my ass here!).

### No parsed matches

First I'd like to discuss the case where there are no parsed matches at all. 

The non-separated case: if we assume each NULL takes up 8 bytes of storage (in the end they should all end up as some kind of 64-bit pointer, right?), and we some day import 6.5 billion player matches, the NULLS for the non-separated case will use 8 * 32 * 6.5 billion = 1.664 terabytes.

The separated case is very simple. Since parsed_player_matches will point back to player_matches iff it exists, having no parsed_player_matches means the cost is 0.

Ok, so with no parsed matches, the separated case wins. That was to be expected.

### All matches parsed

So let's look at the other edge case. We find some kind person who has saved every dota 2 replay, so we set about parsing all of those. The total parsed data will be, well, freaking enormous, but it will be the same for both cases, so I will ignore it.

In this scenario, the extra data for the non-separated case will be 0.

For the separated case, it will instead be the cost of a pointer from each parsed_player_matches back to the associated player_matches. This would use 8 * 6.5 billion = 0.052 terabytes.

52 GB of extra data for this case seems like very little in comparison to the amount of data that would probably have to be stored anyway if all matches were parsed.

### Current numbers

For the current numbers you gave, the same calculation yields:
Non-separated: (760m - 2.4m) * 8 * 32 = 194 GB extra NULL pointers
Separated: 2.4m * 8 = 0.0192 GB (192 MB) extra primary key data for the separate table.

### Realistic future numbers

I expect that the amount of unparsed matches will grow faster than the amount of parsed matches for the foreseeable future. In other words, the extra data in the non-separated case will grow a lot more than the extra data would in the separated case.

I don't expect someone to hand over replays for past matches, so I want to point out that for almost all matches all of those existing NULLs in the non-separated case will stay as NULLs, taking up unnecessary space.

### Conclusions

Even if my assumptions are wrong by a magnitude of 100 (which it probably is, I expect the database to be smarter than to use 32 full-size 64 bit pointers), it seems the separated case wins out in terms of storage use.

## Costs of doing joins

A relational database is basically all about handling relations. In other words: the cost of having to do extra joins to access parsed_player_matches is probably negligible. I can't even begin to speculate in the actual costs.

It should be noted that having a bigger database will probably increase costs for every operation, but this is also something databases are optimized for, so I don't think that's something we need to factor into this.

## Final conclusion (TL;DR)

It seems to make sense to have a separate parsed_player_matches table. But I'm by no means an expert on this.

howardchung:
Yeah, it's everything after additional_units.

Skill is also obtained at a separate time from the initial match data.  Should that go in its own table as well?

JSON vs JSONB: JSON takes less disk space, and I don't plan on querying within the JSON too much.

Part of why I'm doing it the way it is now is to minimize the amount of refactoring that needs to be done to get it compatible with the aggregation framework/front-end templates.  The parsed data usually is an array of events (kills, item purchases, etc.), or a hash of counts (damage dealth, units killed, etc.).

I suppose it's possible to decompose the hash counts into rows in a table, though.

It sounds like both of you recommend splitting off the parsed data into a separate table.  Next question--does this also apply for the match-level data (chat, objectives?)  There are significantly fewer fields for this case.

howardchung:
@gortiz If we're going to put one parsed field into a separate table, then it probably makes sense to do them all.  Would the impact of doing 30+ joins on a query lead to performance issues (or a really messy query that's difficult to maintain?)

howardchung:
@ahockersten according to this it sounds like it would be 8 bytes per row:
http://stackoverflow.com/questions/4229805/how-much-disk-space-is-needed-to-store-a-null-value-using-postgresql-db

ahockersten:
@howardchung Nice link. I'll just update my text a bit with more correct data then. I think the conclusion should still hold, but the numbers will be a lot more manageable for the other case.

howardchung:
Another concern is the cost of adding columns (ALTER TABLE).  Say we want to start tracking camp stacks, and need to put a new column in player_matches.  Would it be very expensive to do this for 6.5 billion rows (all player matches) vs 25 million rows (all parsed_player_matches)?

ahockersten:
Regarding ALTER TABLE, this part from the PostgreSQL docs seems relevant:
"Tip: Adding a column with a default requires updating each row of the table (to store the new column value). However, if no default is specified, PostgreSQL is able to avoid the physical update. So if you intend to fill the column with mostly nondefault values, it's best to add the column with no default, insert the correct values using UPDATE, and then add any desired default as described below."
http://www.postgresql.org/docs/9.5/static/ddl-alter.html

So running an ALTER TABLE would in the worst case scenario incur the cost of updating all 6.5 billion rows. I haven't been able to find any data relating exactly how expensive this is, but I have found numerous people asking how to speed up their ALTER TABLE operations on *millions* of rows, so "really expensive" seems likely :(

howardchung:
Well it would be NULL for previous matches so we wouldn't need to write to
every row?

ahockersten:
:blush: Oh yeah... That's entirely correct.

gortiz:
As you found, PostgreSQL is very efficient when null columns are stored and
very fast when new columns with null value are created. So this is a point
to the "non-separated" version.

On the other hand, I think it will easy to read and more efficient in terms
of RAM memory if we split a huge table in several smaller tables. At least
if YASP does not read the whole "match graph" (I mean: it does not join
when it is not necessary, for example, if parsed data is not necessary to
show information in the main user page, where last games are shown). But
you shouldn't create one table per column. There are two reasons:

   - Some of the columns will be used at the same time, for example *damage*,
   *damage_taken*,* damage_inflictor.*
   - PostgreSQL, as all SQL databases, has a planner that analyze the query
   trying to find the better way to evaluate it. But that is costly operation
   that grows exponentialy with the number of tables, so when the query
   contains several joins (i dont remember exactly how, it is something like
   10), the planner changes its algorithm and it uses an evolutionary
   algorithm which is faster to execute but it usually produce plans that are
   slower to run than the exponential algorithm used when not too many tables
   are involved.
      - TD;LR: Do not join more than 8-10 tables



2015-10-21 13:34 GMT+02:00 Anders Höckersten <notifications@github.com>:

howardchung:
The current model is to compute/aggregate ALL of the data when a player visits any player tab.  We then cache it for two weeks in Redis, incrementally updating it when we add a new match.

Pros:
* Fast loads after the cache build (no db access required after the first initial slow load)
* Simple logic (either use cache or build the profile).  We don't need to write separate queries for each tab.

Cons:
* Memory usage.  We're currently storing roughly 21k cached player profiles and this consumes 8GB of memory.  It doesn't scale very well since RAM is expensive.  We could probably store it with postgres instead (player_caches table that just has two columns, account_id and json cache)
* slow initial load (can take several seconds to generate the page at first)

howardchung:
Deployment pushed back another week.  @albertcui is going to be away the weekend of the 30th.

mhartkorn:
I would change some things about the SQL schema.

Because the match data is static for already parsed matches and only changes for new ones (or do you want to save replays for later parsing?) the data could be held redundant for faster access (but it costs some disk space). Operations like `COUNT()` are very expensive and slow the whole database. What I would do is adding some of meta data to the player profile, to be more specific the match and "replays parsed" counter, the win-lost ratio, and the MMRs (maybe more?). You can save a lot of database queries and aggregation functions with that.

Also the front page is pretty expensive to generate right now. As far as I understand, yasp is currently generating the page once and then saves it in redis. Maybe it is better to go Dotabuff's way and only show the best/most played 10 heroes (instead of all) and keep a JSON about them and the last 10 matches in redis. When a match with a matching hero is played the values in the saved JSON are updated without aggregating the data from the database. There are some edge-cases where this is not very trivial (someone has the same amount of matches with his 10th and 11th best hero) but I think that this is manageable and does apply in <1% of the cases. Because a match cannot be "unplayed" the data should be easy to update and to keep consistent. Downside: A method to check the database for inconsistencies needs to be implemented.

I would store the match data yasp can get from the Dota API and the parsed match data in separate tables because, as @howardchung said earlier, less than 5% of the matches are analyzed at the moment and even if it was 20% it is worth the effort, in my opinion. The joins are only performed when somebody looks at a specific match and the regular profile page does not need it. I also thought about multiplying the columns by 10 and keeping the row count down that way. Could be harder to aggregate data from it, though. But the data for the profile page would be in the other anyways, so not loss there, I think.

As far as match parsing goes, I don't have any optimizations, I only thought about keeping the aggregated data (for e.g. the gold graph) in redis for about 15-60 minutes instead of the whole match data. I don't know the user behavior of yasp users but I don't think that a lot of people are constantly looking at their analyzed matches. Maybe someone can give a few stats?

howardchung:
Re: Count(), we actually barely use it (really only in the status page) and there it should be able to use the index to count so it doesn't need to do a full table scan.

All of the player page aggregations are actually done via JavaScript.  We just have the database engine return an array of matches and do all the counting with JS code.  Partly this was because I didn't really understand MongoDB aggregation and partly because I wanted to do some complicated stuff (like querying/counting teammates/opponents).  Summing together positional data like wards and things like word counts is also simpler with JS than to try to do via DB query.

The redis cache is basically our denormalization/performance improver.  The approach you described for top 10 heroes unfortunately doesn't really work though.  To store an accurate count of heroes played you MUST:

* Include all matches in the count so you have an accurate top 10
* Maintain a record for every hero.  Otherwise heroes not in the top 10 will never have a chance to enter it, since they won't show up in the cached 10 you're saving.

Consistency check is currently implemented via a DB query to count the number of matches the player has, and comparing to the count in the Redis cache.  If they fail to match, we rebuild the cache (expensive DB query).

Re: separate tables.  The joins also need to be performed when building player pages (since we need to join with parsed data in order to aggregate it, like lifetime multikills/killstreaks or wardmap, wordcounts, etc.).  I'm actually thinking of keeping it single-table (less code).  If we have performance issues we could move it later.

Re: dumping old parsed data.  I don't really want to do this.  Yes, it is true that player don't really look at old matches very often.  However, we do have to keep the data around for things like records, wordcounts, wards (basically all the aggregations).

It would also be weird if players decided to check an old match and found that chunks of parsed data were missing.  It should probably be an "all-or-nothing" thing.  I've discussed with @albertcui about expiring parsed data after a while, if storage costs start becoming a problem.

howardchung:
Doing some reading about MVCC/VACUUM/COUNT:
https://wiki.postgresql.org/wiki/Introduction_to_VACUUM,_ANALYZE,_EXPLAIN,_and_COUNT

So it turns out that there is another advantage to storing parsed data in a separate table.  Due to MVCC an update operation just writes a new row and lets the old row be cleaned up later.  It also means the index cannot be used for COUNT(*) purposes since visibility information is not stored there.

PostgreSQL 8.2 added an optimization that checks index pages for no updates/deletes and allows use of the index for COUNT(*) to avoid the full table scan.

If we stored parsed data in a separate table both tables would essentially only contain inserts, no deletes, and very few updates (only when players manually request matches that already exist/are parsed).

Seeing as the migration is almost complete and splitting tables would be pretty costly I probably won't do it unless we have a pressing reason to, but something to keep in mind.

gortiz:
You have a very good point! meanwhile, if it is a performance problem, it
could be better to store that count(*) value as a column on a player table,
but i think we talk about that some weeks ago.

2015-11-11 11:15 GMT+01:00 Howard Chung <notifications@github.com>:

PR #691, opened 2015-10-01
try starting retriever when 80% of accounts are ready

albertcui:


retriever.js
```diff
@@ -100,14 +107,23 @@ async.each(a, function(i, cb) {
...
 
     function allDone() {
-        if (dotaReady && relationshipReady && !launched) {
+        if (dotaReady && relationshipReady) {
             count += 1;
-            console.log("acct %s ready, %s/%s", i, count, users.length);
+            console.log("acct %s ready, %s/%s", i, count, users.length);    
+        }
+        
+        if (!launched) {
```
howardchung: I think you need to check all three conditions before cb()

howardchung: Also this will prevent the last 20% of accounts from cb()?

albertcui: The last 20% of accounts will cb() from the top if statement.

howardchung: i guess.  I feel that this is kind of messy.  Maybe we should make logging in and launching separate things

retriever.js
```diff
@@ -12,13 +12,68 @@ var steamObj = {};
...
+        error: err
+    });
+});
+
+var server = app.listen(port, function() {
+    var host = server.address().address;
+    console.log('[RETRIEVER] listening at http://%s:%s', host, port);
+});
+    
 async.each(a, function(i, cb) {
```
howardchung: does this need to be async if you don't have a callback?  Is the callback required if you're using async?

albertcui: It is not required, and async is better :P

howardchung: how is it any better?

PR #696, opened 2015-10-02
Unparsed

albertcui:


deprecated/unparsed.js
```diff
@@ -1,25 +1,23 @@
...
+    "start_time": {
+        $gt: moment().subtract(1, 'days').format('X')
+    },
+    "parse_status": 0
+}, function(err, docs) {
+    if (err) {
+        return;
+    }
+    async.eachSeries(docs, function(match, cb) {
+        insertMatch(match, function(err, job) {
```
howardchung: You need to queueReq so Kue tries to do the parses again.  Actually i think insertMatch calls queueReq but it's easier to just skip the upsert

deprecated/unparsed.js
```diff
@@ -1,25 +1,23 @@
...
-            queueReq("api_details", match, function(err, job) {
-                console.log("[UNPARSED] match %s, jobid %s", match.match_id, job.id);
-                cb(err);
-            });
-        }, function(err) {
-            console.log(docs.length);
-            done(err, docs.length);
+
+db.matches.find({
+    "start_time": {
```
howardchung: No index on start_time means this will be really slow.  You can try something like match_id

PR #704, opened 2015-10-07
Improvements to quickstart instructions

ahockersten:
I went through the process of building and running YASP on my Fedora machine. This patch adds a new init script to make it easier for anyone who tries the same thing after me. I also thought the instructions were a bit unclear in places

This also fixes an issue in the init script for Debian/Ubuntu. Since contextify (and possibly other dependencies) don't currently run under Node 4.0.0 which has recently been declared the stable version of Node (see https://github.com/brianmcd/contextify/issues/180).

For a more complete list of changes, see the comment to the commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,13 +65,15 @@ Tech
 
 Quickstart
 ----
-* Install dependencies.  If on Debian/Ubuntu: `sudo bash init.sh`  Otherwise, you're responsible for figuring out how to install dependencies yourself.
+* Install dependencies: `sudo bash init.sh`. For other platforms, please have a look at the [wiki](https://github.com/yasp-dota/yasp/wiki/Installation-for-other-platforms).
 * Create .env file with required config values in KEY=VALUE format (see config.js for a full listing of options) `cp .env_example .env`
   * Note: If you have Steam Guard activated on your account you will
     either have to deactivate it or create a new account for use with
     the retriever (recommended).
 * Build `npm run build`
-* Run all services in dev mode (this will run under nodemon so file changes automatically restart the server): `npm run dev`.  You can also start individual services: `npm run dev web,parser`
+* Run `npm test` to make sure your install works correctly
+* Run all services in dev mode (this will run under nodemon so file changes automatically restart the server): `npm run dev`. You can also start individual services.
+
 
 Sample Data
 ----
--- a/config.js
+++ b/config.js
@@ -7,7 +7,7 @@ catch(e){
 var defaults = {
     "STEAM_API_KEY": "", //for API reqs, in worker
     "STEAM_USER": "", //for getting replay salt/profile data, in retriever
-    "STEAM_PASS": "",
+    "STEAM_PASS": "", //make sure to wrap in double quotes if it contains special characters
     "RECAPTCHA_PUBLIC_KEY": "", //for preventing automated requests, in web
     "RECAPTCHA_SECRET_KEY": "",
     "PAYPAL_ID": "", //for donations, in web


howardchung:
* A building version with 4.1.2 is in the `sql` branch, which should be merged into master later this month.  I avoid the issue by not using npm/webpack to build some of the client-side JS.

* mocha, foreman, and nodemon are all development tools and don't need to be installed for production deployments, which is why they're not in the init script (since we use that to bootstrap new instances).

* We may want to put the init scripts in their own subfolder in order to avoid cluttering the top level too much.

* Are you sure Steam Guard is defaulted on for new accounts?  I don't think that has been the case for us (at least I haven't been seeing a lot of errors for new accounts).

* npm script--okay

* quickstart change--okay

config.js
```diff
@@ -7,7 +7,7 @@ catch(e){
 var defaults = {
     "STEAM_API_KEY": "", //for API reqs, in worker
     "STEAM_USER": "", //for getting replay salt/profile data, in retriever
-    "STEAM_PASS": "",
+    "STEAM_PASS": "", //make sure to wrap in quotes if it contains special characters
```
howardchung: dotenv may not like the quotes.  I haven't tested this

README.md
```diff
@@ -65,13 +65,17 @@ Tech
 
 Quickstart
 ----
-* Install dependencies.  If on Debian/Ubuntu: `sudo bash init.sh`  Otherwise, you're responsible for figuring out how to install dependencies yourself.
+* Install dependencies.
```
howardchung: You may just want to want to write a separate Fedora section.

howardchung:
Also note that with #582 we are moving from MongoDB to PostgreSQL.  I won't be able to update the Fedora script to reflect that.

ahockersten:
Wow, that was a quick reply! :)

Since SQL is becoming the default, I suggest that I'll just remake my patch on top of the sql branch. As you say, when #582 is merged I'll have to do that anyway, as installing PostgreSQL requires a bit more hand-holding on Fedora than it does on Ubuntu.

I suggest that I create a specific init script for production servers (name suggestions: init_production.sh, init_dev_debian_ubuntu.sh, init_dev_fedora.sh). I don't think anyone will want to run servers on Fedora anyway, and if they do then my script probably won't cut it. Then the init scripts for dev work can install development dependencies as well. I'll place these in a subfolder as you suggested (I'll call the subfolder "env", unless you want something different. "init" or "environment" might be other good candidates).

You're right, Steam Guard does not appear to be on by default (https://support.steampowered.com/kb_article.php?ref=4020-ALZM-5519#how):
"Steam Guard is enabled by default on your Steam account if your email is verified and you have restarted Steam twice since verifying your email."
I suggest we keep my comment, but I'll modify it to say something along the lines of "Note that Steam Guard may get enabled for new accounts under some circumstances. If it does, you will have to disable it to be able to use the account with YASP".

config.js
ahockersten: The reason I added this comment is that my autogenerated password contained a '#', so if I don't put it inside quotes it will only use half the password and regard the rest of it as a comment.

To be sure, I now double-checked dotenv's github and the readme says "If you need multiline variables, for example private keys, you can double quote strings and use the \n character for newlines:", so enclosing values in double quotes should be safe.

README.md
ahockersten: I considered that, but hesitated. You're right though, it makes sense to do that, especially if someone comes along and adds install instructions for some other platform in the future.

config.js
howardchung: Okay.  I remember having some trouble with quotes before, but I'll bring it up again if we encounter it in production.

howardchung:
Okay, but watch out for merge conflicts.  I just made some changes to the README like 30 seconds ago.

Sure, put them all under some folder like "init" and update the README to match.  I'm not sure it's really necessary to create separate scripts for dev/production.  The difference is one line for the npm dev tools, and those are actually already installed already.  They're just in the node_modules folder so the user either needs to add to their PATH to use them on command line, or add a npm script.

I think the current text that advises users to make sure Steam Guard is disabled is sufficient.

ahockersten:
Hmm, I expected to be able to change the base for this pull request to sql, not to create a merge request for the sql branch + this branch, I'll change the base once I figure out github's UI for it (or create a new pull request, if that's what's required).

The changed commit should include everything we've discussed + one more change that I had to make: Fedora apparently doesn't give global read access to home folders by default, so I changed the "npm run create" target to pipe the contents of sql/create.sql to the psql command instead of reading the file directly. I can't think of a situation where this would change the behavior of it.

ahockersten:
Apparently you can't change the base (http://stackoverflow.com/questions/10081053/how-to-change-the-base-branch-of-a-pull-request). Let's just keep the discussion here and I'll create a new pull request once the sql branch has been merged to master? I'll probably want to double-check that it still works once the sql branch is merged anyway.

howardchung:
Needs to be rebased.  However, I'm thinking that you could just put Fedora-specific instructions in the wiki (someone has done that already with Mac OSX), and just update the README to point there.

For deployment purposes we aren't going to be using Fedora so I'm not sure if it's necessary to check the code/scripts in.

ahockersten:
I moved the instructions to the wiki and updated the README.

init/fedora.sh
```diff
@@ -0,0 +1,10 @@
+sudo dnf check-update && sudo dnf update
```
howardchung: just put the contents in the README?

package.json
```diff
@@ -10,6 +10,7 @@
     "test": "rm -rf testfiles && git clone https://github.com/yasp-dota/testfiles && npm run build && istanbul cover _mocha --report lcovonly -- -R spec",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "constants": "node constantsUpdate.js",
+    "redis": "redis-server",
```
howardchung: I don't think this is necessary.  Generally the advantage of putting things is that node_modules is included in the PATH.  Since redis is not installed in node-modules the user can just run it from the command-line

package.json
```diff
@@ -17,8 +18,8 @@
     "maven": "mvn -q -f java_parser/pom.xml clean install -U",
     "build": "npm install && npm run maven && npm run webpack",
     "startup": "sudo pm2 startup ubuntu -u yasp",
-    "create": "bash sql/create.sh",
     "deploy-core": "git pull origin master && npm run build && node deploy core",
+    "create": "sudo -u postgres psql -f sql/create.sql",
```
howardchung: why does this need to be changed?

ahockersten: Oops. That was a mistake in my rebase. I've also moved the Fedora instructions to the wiki.

README.md
```diff
@@ -65,13 +65,15 @@ Tech
...
+* Install dependencies: `sudo bash init.sh`. For other platforms, please have a look at the [wiki](https://github.com/yasp-dota/yasp/wiki/Installation-for-other-platforms).
 * Create .env file with required config values in KEY=VALUE format (see config.js for a full listing of options) `cp .env_example .env`
   * Note: If you have Steam Guard activated on your account you will
     either have to deactivate it or create a new account for use with
     the retriever (recommended).
 * Build `npm run build`
-* Run all services in dev mode (this will run under nodemon so file changes automatically restart the server): `npm run dev`.  You can also start individual services: `npm run dev web,parser`
+* Run `npm test` to make sure your install works correctly
+* Run all services in dev mode (this will run under nodemon so file changes automatically restart the server): `npm run dev`. You can also start individual services.
+
```
howardchung: put the `npm run dev web,parser` example in

albertcui:
Great, thanks for the contribution! I'll update the individual services section myself.

PR #705, opened 2015-10-07
Cache test repository between test runs

ahockersten:
Pretty self-explanatory. I got tired of downloading the full test repository every time I wanted to run the test suite.

From what I've been able to gather the files inside the "testfiles" directory are not actually changed by the testsuite. If that's the case it would be possible to make this patch a lot simpler by just reusing the same repo and doing "git pull" inside it, but it seems like good practice to start with a clean repo for the testfiles anyway.

This will end up using a bit more disk space than previously, but using a bare (--mirrored implies --bare) repository for "testfiles_cache" should keep that at a minimum, at least on platforms where git supports hard linking.


howardchung:
Usually I just use `mocha` to run the tests without redownloading the test files.

ahockersten:
But then you have two commands which both can be used to run the "same" tests. In my experience, that may cause problems down the line. If it was my decision, I would run every part of the build process and testing (or at least as much as possible) through different npm commands.

On a different but related note, I have submitted two pull requests to you today, and I should probably introduce myself a bit and explain what I'm up to. I really like YASP, and I thought I should contribute somehow. The first step to contributing to any project is to get it to build, and while doing that I noticed some stuff that I thought could be handled better. I might as well submit some patches and see what you think about them to get over the "hump" of submitting patches to the project. Hopefully, somewhere down the line (tomorrow, if I have as much time available as I hope to have) I will think of or find a feature or a bug in the issues list that I want to implement, and then I'll do that. 

Finally, as you very well know, you own the project, and you have to be the final arbiter on what goes in and what gets rejected. I won't be sad or upset if you don't merge any of my patches. I realized when I wrote this patch that you might end up rejecting it with an explanation like the one you just gave, and I'm very ok with that response. :)

howardchung:
The only part I don't like is having two testfiles directories.  Also it would be nice to keep the npm scripts relatively short.

howardchung:
What problems do you envision?  mocha is supposed to run the tests.  npm test just grabs the files first, then does mocha.

ahockersten:
Hmm, I apparently wrote most of this comment before going to bed and then pushed my change without clicking the "comment" button.

I moved the testfiles directories to the test subdirectory. I also moved all of the commands into a bash scripts in the test subdirectory.

I don't envision any specific problems with the current setup. If for whatever reason a new step is required in the future, then "npm test" and "mocha" will do different things, and at that point it's good to be in the habit to always run the tests the "right" way (which should be `npm test` rather than `mocha` right?).

I do find it potentially confusing for newcomers - before you pointed it out to me I had no idea that running `mocha` and `npm test` have the same effect.

howardchung:
mocha is still going to run the tests.  I don't think it's any more "right" or "wrong" to use `npm test`.  The idea is, npm test will bootstrap the whole process by downloading replay files.  It will always work.

If you want to run a specific test, or you know you don't need to download replays again, you can use `mocha`.

That said this PR is probably okay to merge now. @albertcui can decide if he likes it.

howardchung:
@ahockersten if you're still interested in merging this please resolve the conflicts

ahockersten:
Conflicts resolved

howardchung:
Travis doesn't like it.  Can you debug?

test/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+rm -rf test/testfiles
+if [ ! -d "test/testfiles_mirror" ]; then
+  git clone --mirror https://github.com/yasp-dota/testfiles test/testfiles_mirror
```
howardchung: @ahockersten remind me what the purpose of the mirror is again?  Can't we just pull into testfiles (which is already a repo?)

The reason I did the deletion every time was for idempotence (so the test could be run multiple times in dev environment, since cloning into an existing repo errors).

ahockersten: That was overengineering it a bit, now that I look at it again. I have updated to use clean together with a pull instead. That should be exactly the same as removing it.

test/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+if [ ! -d "test/testfiles" ]; then
+  cd test
+  git clone https://github.com/yasp-dota/testfiles
+  cd ..
+else
+  cd test/testfiles
+  git clean -fxd
```
howardchung: won't just a regular git pull work?

howardchung:

albertcui:
Looks good. Thanks!

PR #707, opened 2015-10-08
Adds visual space for empty item slot #702

### LINKED ISSUE #702: Items in inventory are visually collapsed
For instance, my friend when he builds force staff, will always keep it in his 4th item slot.

In example: [Match 1848421474](http://yasp.co/matches/1848421474), he plays the Shadow Shaman. The TP boots are in item slot 1, so that looks correct. However the force staff is in the 4th item slot and the euls is in the 5th item slot.

So, we'd both like some spacers filling in the empty inventory slots. However, I don't know if the inventory slot index is currently tracked?

ahockersten:
fixes #702

### DIFF
--- a/views/match/match_index.jade
+++ b/views/match/match_index.jade
@@ -70,7 +70,7 @@ block match_content
                   if (index.slice(0, 5)==="item_")
                     - var item = constants.items[constants.item_ids[val]]                             
                     if item
-                      img.img-small.ability(src=item.img, title=constants.item_ids[val])
+                      img.img-small.item(src=item.img, title=constants.item_ids[val])
                     else
                       span.spacer-small.item
 
--- a/views/match/match_index.jade
+++ b/views/match/match_index.jade
@@ -62,6 +62,8 @@ block match_content
                   - var item = constants.items[constants.item_ids[val]]
                   if item
                     img.img-small.item(src=item.img, title=constants.item_ids[val])
+                  else
+                    span.spacer-small.item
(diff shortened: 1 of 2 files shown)


howardchung:
Looks okay, but number may need to be tweaked later (I changed some sizes in sql branch)

howardchung:
@albertcui please review

views/match/match_index.jade
```diff
@@ -62,6 +62,8 @@ block match_content
                   - var item = constants.items[constants.item_ids[val]]
                   if item
                     img.img-small.item(src=item.img, title=constants.item_ids[val])
+                  else
+                    span.spacer-small.item
```
howardchung: This should also be done for lone druids bear

ahockersten: Nice catch. I'm fixing this. However, I found another bug while doing so (reported as #709). I'll look into fixing that one as well as a part of this patch.

views/match/match_index.jade
```diff
@@ -69,7 +71,9 @@ block match_content
                     - var item = constants.items[constants.item_ids[val]]                             
                     if item
                       img.img-small.ability(src=item.img, title=constants.item_ids[val])
```
howardchung: Found the issue.  .ability should be .item

albertcui:
I don't really like the spacer. I don't think which item slots are empty is useful information, and since Dotabuff also has what we're doing, I think it would be confusing for users to add the spacer.

I've added the commit for fixing #709 though. Thanks!

PR #710, opened 2015-10-09
Reloads/resizes heatmap on map resize. Fixes #605

### LINKED ISSUE #605: Heat Map issues (browser compatibility, sometimes requires refreshes to load)
The heat map on http://yasp.co/matches/xxxxxxxx/positions works in other browsers, but not Safari.

ahockersten:
fixes #605 

- Also fixes similar issues for player ward placement data

### DIFF
--- a/public/js/resize.js
+++ b/public/js/resize.js
@@ -1,0 +1,157 @@
+// resize detection with some slight changes to fit module system from:
+// http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/
+
+var attachEvent = document.attachEvent;
+var isIE = navigator.userAgent.match(/Trident/);
+var requestFrame = (function(){
+  var raf = window.requestAnimationFrame || window.mozRequestAnimationFrame || window.webkitRequestAnimationFrame ||
+      function(fn){ return window.setTimeout(fn, 20); };
+  return function(fn){ return raf(fn); };
+})();
+
+var cancelFrame = (function() {
+  var cancel = window.cancelAnimationFrame || window.mozCancelAnimationFrame || window.webkitCancelAnimationFrame ||
+         window.clearTimeout;
+  return function(id){ return cancel(id); };
+})();
+
+function resizeListener(e) {
+  var win = e.target || e.srcElement;
+  if (win.__resizeRAF__) cancelFrame(win.__resizeRAF__);
+  win.__resizeRAF__ = requestFrame(function(){
+    var trigger = win.__resizeTrigger__;
+    trigger.__resizeListeners__.forEach(function(fn){
+      fn.call(trigger, e);
+    });
+  });
+}
+
+function objectLoad(e) {
+  this.contentDocument.defaultView.__resizeTrigger__ = this.__resizeElement__;
+  this.contentDocument.defaultView.addEventListener('resize', resizeListener);
+}
+
+window.resize = {};
+
+window.resize.addResizeListener = function(element, fn) {
+  if (!element.__resizeListeners__) {
+    element.__resizeListeners__ = [];
+    if (attachEvent) {
+      element.__resizeTrigger__ = element;
+      element.attachEvent('onresize', resizeListener);
+    }
+    else {
+      if (getComputedStyle(element).position == 'static') element.style.position = 'relative';
+      var obj = element.__resizeTrigger__ = document.createElement('object');
+      obj.setAttribute('style', 'display: block; position: absolute; top: 0; left: 0; height: 100%; width: 100%; overflow: hidden; pointer-events: none; z-index: -1;');
+      obj.__resizeElement__ = element;
+      obj.onload = objectLoad;
+      obj.type = 'text/html';
+      if (isIE) element.appendChild(obj);
+      obj.data = 'about:blank';
+      if (!isIE) element.appendChild(obj);
+    }
+  }
+  element.__resizeListeners__.push(fn);
+};
+
+window.resize.removeResizeListener = function(element, fn) {
+  element.__resizeListeners__.splice(element.__resizeListeners__.indexOf(fn), 1);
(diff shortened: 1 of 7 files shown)


ahockersten:
I haven't tested this on mobile yet, because I expect to change it based on feedback. It works on desktop though.

ahockersten:
Before anyone asks about it: I decided to copy the contents of resize.js from some blog, because there really wasn't any cleaner way that I could find. I found lots of half-baked libs on github for solving this issue. Some of them were even based on this code, but none that seemed to be available in a finished form in NPM, and were under active maintenance. In the end, this seemed like the least bad solution.

public/js/resize.js
```diff
@@ -0,0 +1,65 @@
...
+      obj.type = 'text/html';
+      if (isIE) element.appendChild(obj);
+      obj.data = 'about:blank';
+      if (!isIE) element.appendChild(obj);
+    }
+  }
+  element.__resizeListeners__.push(fn);
+};
+
+module.exports.removeResizeListener = function(element, fn){
```
howardchung: note that I'm not using webpack in the current version of `sql` to avoid build issues.  I'm just making each of these files a window.functionName = function(){}.

howardchung: It should work to just replace module.exports.functionName with window.functionName.

On the webpack side you can just do require() and in base_layout.jade you can just write a script tag that includes this file.  Check out the current code for examples.  The idea is to be able to toggle between webpack and non-webpack easily.

ahockersten: I don't think I understand your comment. I can't replace module.exports.functionName() with window.functionName() and still call window.resize.functionName() the way I currently do? Do you want me to rebase the pull request on the sql branch instead, where such a rewrite makes sense?

howardchung: You could have resize JS assign an object to window with those functions.

In general yes, you should do feature development on sql branch to avoid
merge conflicts later.

ahockersten: I'll rebase this once sql gets merged to master then otherwise I'll probably have to rebase it again once it gets merged anyway.

howardchung: Okay.  Thanks for the work, I don't think @albertcui or I would have gotten to this.

You could pull request into `sql` branch if you'd like earlier integration.

ahockersten: I think I'll wait actually. In my opinion, the less work going in on the sql branch the better and this is hardly a critical fix.

albertcui:
Just to be clear, we're going to merge this into sql instead of master?

howardchung:
I think @ahockersten was planning to wait until after the sql->master merge, but I think it also works to have it merged into sql directly.

ahockersten:
As I said above, the less stuff unrelated to sql going in on the sql branch the better, in my opinion. Branches should be kept as small and topical as possible. There will likely be regressions from the merge of the sql branch (yes, I'm a pessimist when it comes to code :)), and there will be enough trouble chasing down those without having to look at totally unrelated things.

howardchung:
This is fully client-side, so I think it's probably okay.  It's up to you :)

There will almost certainly be some issues, whether it's bugs we can't find in dev or just trying to migrate 75 million matches. :P

howardchung:
@ahockersten if you want to rebase this go ahead.  However I've gotten rid of webpack so your include of resize.js will need to go in the base_layout.jade

ahockersten:
I've updated resizing and tested it on all heatmaps (including the new "performances").

deprecated/match_positions.jade
```diff
@@ -27,20 +27,45 @@ block match_content
 
 append footer_assets
   script.
-    var width = document.getElementById('map').clientWidth;
-    var heatmap = h337.create({
-      container: document.getElementById('map'),
-      radius: 15*(width/600)
-    });
+    (function() {
```
howardchung: I wouldn't bother updating this since it's deprecated

views/match/match_performances.jade
```diff
@@ -80,20 +80,51 @@ block match_content
                   
 append footer_assets
   script.
+    (function() {
```
howardchung: should probably refactor this since it's used twice (performances/vision)

views/player/player_wardmap.jade
```diff
@@ -3,28 +3,75 @@ extend player
...
-    });
-    var heatmap2 = h337.create({
-      container: document.getElementById('map2'),
-      radius: 15
-    });
-    heatmap.setData(posData[0]["obs"]);
-    heatmap2.setData(posData[0]["sen"]);
-          
\ No newline at end of file
+    (function() {
```
howardchung: same refactoring?  This one's a bit different since two maps but should be doable

albertcui:
Looks good for merge after duplicated code is refactored.

views/match/match_performances.jade
```diff
@@ -80,20 +80,51 @@ block match_content
...
+      }
+
+      function mapResized() {
+        var width = $('#map').outerWidth();
+        // since h337 overlays the map with its own element, we need to
+        // recreate this element so that the resize works as intended
+        if (heatmap) {
+          $(heatmap._renderer.canvas).remove();
+        }
+        heatmap = h337.create({
```
howardchung: also duplication here in numbers within the function and outside.

ahockersten:
Refactoring done.

ahockersten:
I moved stuff around a bit more than requested. Since adjustHeatmapData was only used by our heatmaps I moved it inside resize.js. I also made it non-destructive so I could reuse the posData in an easier way.

howardchung:
Push the commit?

ahockersten:
Finally pushed, after resolving some conflicts with current master

howardchung:
Awesome, thanks. @albertcui please review

albertcui:
Thanks!

albertcui:
Deployed.

PR #715, opened 2015-10-13
Use yasp db in production and yasp_test in test

ahockersten:


ahockersten:
Tests fail because init.sh needs to be run once for the yasp user to be created. It's an easy fix to change test.js to handle that, but I assumed init.sh was always run at least once. I haven't actually ever used Travis, so I don't know what the expected way to solve stuff like this is.

ahockersten:
Seems like the answer is that Travis doesn't do that (I'm not sure why I expected anything different). I'll push a fix.

ahockersten:
This has a bug that I introduced in the database creation, please don't merge it yet.

howardchung:
Travis just runs npm test.  Can you make that create the database and be
idempotent so we can run it repeatedly in development?

ahockersten:
This is just to confirm that the latest commit works with `npm test`, travis passes and I could load and parse a match when running under `npm run dev`. I removed the previous comment about Travis.

howardchung:
Are both npm test and npm run create idempotent?

I use npm run create to reset the db to clean state so I can test migration
script.

ahockersten:
`npm run create` is not idempotent. I missed that it used to be, but that's an easy fix.

ahockersten:
They should now both be idempotent.

package.json
```diff
@@ -17,7 +17,7 @@
     "maven": "mvn -q -f java_parser/pom.xml clean install -U",
     "build": "npm install && npm run maven && npm run webpack",
     "startup": "sudo pm2 startup ubuntu -u yasp",
-    "create": "sudo -u postgres psql -f sql/create.sql",
+    "create": "sh sql/create.sh",
```
howardchung: bash for consistency?

sql/create.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+sudo -u postgres dropdb yasp
+sudo -u postgres dropuser yasp
+sudo -u postgres createuser yasp
+sudo -u postgres psql -c "ALTER USER yasp WITH PASSWORD 'yasp';"
+sudo -u postgres createdb yasp --owner yasp
+cat sql/create_tables.sql | sudo -u postgres psql -d postgresql://yasp:yasp@localhost/yasp -f -
```
howardchung: why the cat | instead of just using -f directly?

sql/create_tables.sql
```diff
@@ -1,9 +1,3 @@
---postgres user by default connects to postgres db
-ALTER USER postgres WITH PASSWORD 'postgres';
---default schema is public
-drop schema public cascade;
```
howardchung: how are you "resetting" the database to clean state?

sql/create.sh
ahockersten: If using -f directly, the postgres user needs read rights to create_tables.sql.

In my case, yasp is located in /home/anders/git/yasp, and the default of Fedora is (apparently) to not give other users read access to users' home directories.

sql/create_tables.sql
ahockersten: There's calls to `dropdb` and `dropuser` in sql/create.sh for the production db, and similar calls inside test.js for the test db.

package.json
ahockersten: Agreed. Pushing change for this now.

howardchung:
Okay looks good, @albertcui please review

sql/create.sh
howardchung: Why does that matter if you're running under sudo anyway

ahockersten: `psql -d` is run with sudo under user postgres, so it's running with the access of user postgres, not root.

howardchung: ah okay

On Fri, Oct 30, 2015 at 4:02 AM, Anders Höckersten <notifications@github.com

albertcui:
Awesome, thanks!! I was gonna change this myself and was happy to find you already have :)

PR #723, opened 2015-10-27
Add full date as title to "x days ago" on records page

teuneboon:
Don't even know if you want to have this feature, or if I implemented it correctly(don't have a computer I can test it on right now). Was just familiarizing myself a little bit with the codebase(I like this project and want to make more useful contributions in the future) and found this was easy to add.

p.s: do you guys hang out on some IRC server or something? Would love to help with stuff sometimes as a little side-project :)

### DIFF
--- a/views/player/player_records.jade
+++ b/views/player/player_records.jade
@@ -28,7 +28,7 @@ block playercontent
                     else
                       val.max_match.hero_id
                   td(class=(key in times ? "format-seconds" : ""), data-format-seconds=val.max) #{val.max}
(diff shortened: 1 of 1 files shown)


howardchung:
You can try Gitter: https://gitter.im/yasp-dota/yasp#utm_source=notification&utm_medium=email&utm_campaign=unread-notifications

Email notification seems to be about 1 hour delayed though.  You might get faster response by just opening a GitHub issue and commenting.

Re: your feature.  Sure, but the way we normally add tooltips is with abbr (check match.jade for example).  This might work anyway since I think we have qtip2 automatically work on any "title" attribute--would have to test it.

@albertcui review?

teuneboon:
@howardchung yeah, reason I didn't use abbr here because it looked kind of weird in combination with the <a> here. But it would be a simple change

howardchung:
Did you test this?

PR #738, opened 2015-11-02
Fixed application crash in /status if PostgreSQL connection is lost.

mhartkorn:
yasp crashes if there is no connection to PostgreSQL. This could happen if the SQL server crashes or goes down for maintenance.

The reason is that `count` is then undefined and has no attribute `rows`.

So far the commit displays a white page with a stacktrace. Might not be the most elegant way but the player pages behave the same way. Is there an easy way to make some pretty 500 error pages from the `status` module?


mhartkorn:
I don't know why the Travis-CI check fails. The commit only has 3 lines which run perfectly fine here. `npm test` also works locally.

howardchung:
looks like ENOSPC (you can view the log).  I did a rebuild

howardchung:

albertcui:
Thanks for the contribution!

PR #747, opened 2015-11-12
add endpoint to scanner and workServer, fix case where matches.player…

albertcui:
…s isn't there?

### DIFF
--- a/scanner.js
+++ b/scanner.js
@@ -15,6 +15,29 @@ var constants = require('./constants');
 var retrieverArr = config.RETRIEVER_HOST.split(",");
 var trackedPlayers;
 var userPlayers;
+
+// Used to create endpoint for monitoring
+var express = require('express');
+var moment = require('moment');
+var app = express();
+var port = config.PORT || config.SCANNER_PORT;
+var startedAt = moment();
+
+app.route("/").get(function(res, req) {
+    redis.get("match_seq_num", function(err, result) {
+        res.json({
+            started_at: startedAt.format(),
+            started_ago: moment.fromNow(),
+            match_seq_num: result || "NOT FOUND"
+        })
+    })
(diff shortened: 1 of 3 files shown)


scanner.js
```diff
@@ -15,6 +15,29 @@ var constants = require('./constants');
...
+var app = express();
+var port = config.PORT || config.SCANNER_PORT;
+var startedAt = moment();
+
+app.route("/").get(function(res, req) {
+    redis.get("match_seq_num", function(err, result) {
+        res.json({
+            started_at: startedAt.format(),
+            started_ago: moment.fromNow(),
+            match_seq_num: result || "NOT FOUND"
```
howardchung: scanner supports several modes.  Also won't this not update with the scanner as it scans?

howardchung:
Why do we need this :P

albertcui:
Because this way we know if one of our services are down

howardchung:
How are you going to monitor?  If they stall, then uptime check won't alert because the endpoint will still be up

albertcui:
The stalls are caused by crashes for the most part.

howardchung:
But pm2 will just bring them back up

howardchung:
If they are stalling it probably means we're not calling back somewhere

albertcui:
Doesn't ways happen. Today's work server crashes were not brought back.

howardchung:
If it's constantly crashing and coming back an uptime check may not catch
it.

albertcui:
Yes, but I'm saying today it crashed and never came up. I know because I
manually hit the endpoint.

howardchung:
You deployed this to production?

albertcui:
No I'm saying I manually hit the parse endpoint on work server.

PR #749, opened 2015-11-14
Group

howardchung:
TODO
- [x] Run initial migration
- [ ] deploy new code
- [ ] run second migration
- [ ] clear caches

Implements storing account_id, hero_id, player_slot of each player in a match in a JSON blob as part of the match table (a new column called pgroup).

Requires the use of a migration script to add and fill the pgroup column for all matches.  This needs to be done prior to deploying this change (so the script needs to ported separate to master and run on prod).

@albertcui doesn't really want to do this, since it denormalizes the data somewhat.

The advantage is that we no longer have to query for the peers/heroes played with on team/against separately, which is currently really slow (I'm seeing times of 15-30 seconds but might be improved somewhat by building index on match_id for player_matches).

Don't have to merge, but keeping this open as an option.


howardchung:
Sticky bits:

If we do the migration first, we may not have pgroup for the matches that added while the migration happens.

If we deploy the code first, player profiles could be messed up until the migration completes.

howardchung:
I think we can do the migration first, then deploy new code, then run migration again for the ones we missed during migration.  Just need to modify the migration script to find the matches where !match.pgroup

aggregator.js
```diff
@@ -606,9 +606,9 @@ module.exports = function aggregator(matches, groups, fields, existing) {
             }
         }
         var player_win = isRadiant(m) === m.radiant_win;
-        var group = groups[m.match_id];
-        for (var j = 0; j < group.length; j++) {
-            var tm = group[j];
+        var group = m.pgroup;
```
howardchung: add || {} to protect from crash

PR #752, opened 2015-11-16
added json of notable players and associated account_ids

jford8820:


howardchung:
Awesome, thanks.  It's pretty trivial to start parsing their games, but I'm not sure if it would be very useful if most people don't know about it

jford8820:
Right. To me yasp has a niche as the stats-focused match analysis tool. You could make it content-driven, like 'Comparison of arteezy and burning's major performances' with net worth/time graphs and such but that sounds difficult to automate. How do you see fitting notable players into yasp? I love dota and want to learn so I'm open to whatever

howardchung:
datDota already does a lot of pro-focused analysis.  I'm not sure what else we could add on top of that.  I don't have any ideas, but am open to suggestions.

jford8820:
Item timings are a big part of why I prefer yasp over other sites. Average
'first item' timings by hero for notable players would be cool. It's very
specific but I'd like to know which player has the best blink dagger timing
on tusk in 6.85, for example.

On Sunday, November 15, 2015, Howard Chung <notifications@github.com> wrote:

> datDota already does a lot of pro-focused analysis. I'm not sure what else
> we could add on top of that. I don't have any ideas, but am open to
> suggestions.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/yasp-dota/yasp/pull/752#issuecomment-156888679>.
>


-- 

Jack Ford
Customer Success Engineer - Salsify
(205) 572 - 2429

howardchung:
We have to decide if we just want to build the tools that offer custom analysis (sort of like the filter options on player pages), or pre-created data sets (kind of like the Distribution statistics).

I think that the filter tools are pretty underused, tbh, even though IMO they're pretty cool.  Most people just don't know/care enough to construct the proper queries for interesting data.  Perhaps this is an argument to provide pre-created data.  However, this limits us from providing really specific data (because if you try to compute item timings for every pro player for every patch in advance, you quickly find the computation is infeasible)

howardchung:
Also this has been mentioned before but just "average item timing" as a stat can be easily thrown off based on item build order.  Perhaps just gold earned at x time is a better measure of performance/farming ability.

howardchung:
@albertcui should we merge this first and figure out what we want to do with it later?

jford8820:
Certainly. It would take a variety of assumptions (i.e. adjust/drop the
game if Tusk stops by for mana boots first) to come up with the numbers
although it would be interesting to see.

I understand your decision here. This would be very expensive alongside the
assumptions required to get anything useful. I dug into your implementation
of the Java parser. To my understanding, it doesn't look like there's any
way to make parsing less expensive without forking the parser itself. A
'light' parse would open a lot of doors.

I'll poke around for other things to help out on. Thanks

On Tuesday, November 17, 2015, Howard Chung <notifications@github.com>
wrote:

> Also this has been mentioned before but just "average item timing" as a
> stat can be easily thrown off based on item build order. Perhaps just gold
> earned at x time is a better measure of performance/farming ability.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/yasp-dota/yasp/pull/752#issuecomment-157312672>.
>


-- 

Jack Ford
Customer Success Engineer - Salsify
(205) 572 - 2429

howardchung:
I'm not so worried about the cost of parsing as I am about computing the
averages.

jford8820:
All the more reason to focus my energy elsewhere. Hope the json is useful though

albertcui:
Thanks for the contribution. I think we should just make a "Pro Players" tab or something listing all the pro players we're tracking.

PR #756, opened 2015-11-17
Datefilter - Filter on match starting_time

### LINKED ISSUE #750: date filter for player pages
Requested feature.

Shouldn't be too hard to add (possibly even an entry-level project for a new contributor!)

What options should we offer?  Last week, last month, last 6 months?

WTFKr0:
fixes #750


WTFKr0:
Linked to Issue #750

filter.js
```diff
@@ -23,6 +23,9 @@ module.exports = function filter(matches, groups, filters) {
         lobby_type: function(m, key) {
             return m.lobby_type === key;
         },
+        date: function(m, key) {
+            return m.start_time > Math.floor(Date.now() / 1000) - constants.date[key].duration;
```
howardchung: It might be better to do this once instead of doing it here (this runs for every match)

howardchung: I would drop the duration bit and just filter on start_time

json/date.json
```diff
@@ -0,0 +1,22 @@
+{
```
howardchung: Instead of defining constants you could probably just use moment() in the Jade template to compute the values

howardchung: Perhaps compute the actual unix timestamp and have the filter options give those as values

views/query.jade
```diff
@@ -72,6 +72,11 @@ h2.inline-header
           select.form-control(multiple name="lobby_type" placeholder="Any Lobby")#lobby_type
             each entry in constants.lobby_type
               option(value=entry.id, selected=(q.lobby_type && q.lobby_type[0]===entry.id))=entry.name
+        .form-group
+          label Date
+          select.form-control(multiple name="date" placeholder="Any Date")#date
+            each entry in constants.date
+              option(value=entry.id, selected=(q.lobby_type && q.lobby_type[0]===entry.id))=entry.name
```
howardchung: Actually to allow the query form to "reload" with the selected settings we need to assign ids to each time range (like how you've done).  I don't like that it means the date param loses semantic meaning, but maybe we have to do it this way

howardchung: that said you still have the q.lobby_type equivalency check, you should check for the date param now.

WTFKr0:
Made some changes but don't understand how you want use moment in my case.
Now the date value is a number of days, so you can explicitly add date=14 in url to get last 2 weeks
I let you check

filter.js
```diff
@@ -23,6 +23,9 @@ module.exports = function filter(matches, groups, filters) {
         lobby_type: function(m, key) {
             return m.lobby_type === key;
         },
+        date: function(m, key) {
+            return m.start_time / 86400 > curtime - key;
```
howardchung: don't divide here, just do m.start_time > (curtime - key)

filter.js
```diff
@@ -59,6 +62,7 @@ module.exports = function filter(matches, groups, filters) {
             });
         }
     };
+    var curtime = Math.floor(Date.now() / 86400000);
```
howardchung: just divide by 1000 here to convert JS timestamp to unix timestamp

views/query.jade
```diff
@@ -72,6 +72,12 @@ h2.inline-header
           select.form-control(multiple name="lobby_type" placeholder="Any Lobby")#lobby_type
             each entry in constants.lobby_type
               option(value=entry.id, selected=(q.lobby_type && q.lobby_type[0]===entry.id))=entry.name
+        .form-group
+          label Date
+          select.form-control(multiple name="date" placeholder="Any Date")#date
+            - var date_list = {"7":{"id":"7","name":"Last 1 Week"},"30":{"id":"30","name":"Last 1 Month"},"90":{"id":"90","name":"Last 3 Months"},"180":{"id":"180","name":"Last 6 Months"}}
```
howardchung: use array, not object

WTFKr0:
3 new commits from your recommandations
Kr0

filter.js
```diff
@@ -23,6 +23,9 @@ module.exports = function filter(matches, groups, filters) {
         lobby_type: function(m, key) {
             return m.lobby_type === key;
         },
+        date: function(m, key) {
+            return m.start_time > curtime - (key * 86400);
```
howardchung: may want to parenthesize to ensure order of operations

howardchung:

albertcui:
Looks good! Thanks for the contribution!

PR #757, opened 2015-11-20
Regionfilter

howardchung:


filter.js
```diff
@@ -23,6 +23,11 @@ module.exports = function filter(matches, groups, filters) {
         lobby_type: function(m, key) {
             return m.lobby_type === key;
         },
+        region: function(m, key) {
+          if (constants.regions_id[key] && constants.regions[constants.regions_id[key]] && constants.regions[constants.regions_id[key]].clusters)
+            return constants.regions[constants.regions_id[key]].clusters.indexOf(m.cluster.toString()) != -1;
```
howardchung: I'm a little worried about the performance of indexOf here.  Perhaps we should map clusters to region ids, and give each match a region id (in compute.js).  Then we can just do an equality check here.

constants.js
```diff
@@ -84,6 +87,7 @@ for (var key in regions) {
     }
 }
 cluster["121"] = "US EAST";
```
howardchung: basically replace constants.cluster with a cluster to region id mapping.  Then have compute.js give each match a numerical region ID.  Then we update the views that reference regions to translate the region ID into a friendly string.

PoonpaezYeahYeah: Could you please stop sending the email to me

howardchung: You should be able to "unwatch" the repository to stop getting emails

WTFKr0: Do you know why cluster 121 is not in the region.json ?

howardchung: no, it seems to be a weird exception, although it may have been added since the last time I checked

WTFKr0:
I just change the cluster constant to a id/name array, instead of just name

howardchung:
Check players/count.jade as well.  You may need to change the field name (I think it currently looks for .name but you named it region_name for region)

WTFKr0:
No thats OK cause if .name doesn't exists, it take the value of key : 
td=constants[key] ? (constants[key][key2] ? constants[key][key2].name || constants[key][key2] : JSON.stringify(constants[key])) : key2

On my dev it seems OK too : 
http://dev.yasp.kr0.fr/players/62640399/counts

howardchung:
Caching is on in your box.  It displays the region ID when i forced rebuild
(add a random query string like ?l=0)

WTFKr0:
Yh caching on, I just desactivate it, same result.
Don(t know why your query got this result

WTFKr0:
OK I understand, I look at the problem

howardchung:
We can have region replace cluster.  You just have to name it name instead of region_name in constants.js

WTFKr0:
Not sûre to understand what you mean by : region replace cluster...
I just rename region_name property to region

howardchung:
We can just have constants.cluster be a mapping of cluster ids to region ids.  Then in compute.js we set match.region based on match.cluster.  In the views we translate match.region to a string.

WTFKr0:
"In the views we translate match.region to a string"

Don't know how to do this, do WE need an utility fonction ? The match.region (id) is not any keys of our constants.

howardchung:
You can build the lookup in constants.js (region id to string, or region id to object with id and string)

constants.js
```diff
@@ -77,13 +77,11 @@ var regions = constants.regions;
...
         regions[key].clusters.forEach(function(c) {
-            cluster[c] = regions[key].display_name.slice("#dota_region_".length).split("_").map(function(s) {
-                return s.toUpperCase();
-            }).join(" ");
+            cluster[c] = regions[key].region;
         });
     }
 }
-cluster["121"] = "US EAST";
+cluster["121"] = "2";
```
howardchung: for readability you might want to do regions["USEast"].region

WTFKr0:
Which name tu use for this new constant ? In the counts view, by default the name of the section is the name of the constant use. Don't know if you know what i mean :D

howardchung:
use region

WTFKr0:
OK here : http://dev.yasp.kr0.fr/players/62640399/counts?l=0
But the region name aren't in the same order, comapred to : 
http://yasp.co/players/62640399/counts?l=0

howardchung:
thats fine, order doesnt really matter

views/match/match.jade
```diff
@@ -34,7 +34,7 @@ block content
           tr
             //td= constants.parse_status[match.parse_status]
             td= constants.game_mode[match.game_mode] ? constants.game_mode[match.game_mode].name : match.game_mode
-            td= constants.cluster[match.cluster] ? constants.cluster[match.cluster] : match.cluster
+            td= constants.cluster[match.cluster] ? constants.cluster[match.cluster].region_name : match.cluster
```
howardchung: translate via region

filter.js
```diff
@@ -23,6 +23,9 @@ module.exports = function filter(matches, groups, filters) {
         lobby_type: function(m, key) {
             return m.lobby_type === key;
         },
+        region: function(m, key) {
+          return Number(m.region_id) === key;
```
howardchung: number cast probably not needed

views/query.jade
```diff
@@ -72,6 +72,11 @@ h2.inline-header
           select.form-control(multiple name="lobby_type" placeholder="Any Lobby")#lobby_type
             each entry in constants.lobby_type
               option(value=entry.id, selected=(q.lobby_type && q.lobby_type[0]===entry.id))=entry.name
+        .form-group
+          label Region
+          select.form-control(multiple name="region" placeholder="Any Region")#region
+            each reg, name in constants.regions
```
howardchung: up to you if you want to use constants.regions or constants.region (id to string mapping).  The second might be easier to read

constants.js
```diff
@@ -71,19 +71,22 @@ for (var i = 0; i < 128; i++) {
...
 }
 var cluster = {};
+var region = {};
 //Remove regions nesting
 constants.regions = constants.regions.regions;
 var regions = constants.regions;
 for (var key in regions) {
+    region[regions[key].region] = regions[key].display_name.slice("#dota_region_".length).split("_").map(function(s) {
+      return s.toUpperCase();
+    }).join(" ");;
```
howardchung: double semicolon

constants.js
```diff
@@ -71,19 +71,22 @@ for (var i = 0; i < 128; i++) {
...
 for (var key in regions) {
+    region[regions[key].region] = regions[key].display_name.slice("#dota_region_".length).split("_").map(function(s) {
+      return s.toUpperCase();
+    }).join(" ");
     if (regions[key].clusters) {
         regions[key].clusters.forEach(function(c) {
-            cluster[c] = regions[key].display_name.slice("#dota_region_".length).split("_").map(function(s) {
-                return s.toUpperCase();
-            }).join(" ");
+            cluster[c] = regions[key].region;
```
howardchung: Number() here and you don't have to do it for every match in the filter?

howardchung:
signoff, @albertcui please review (merge conflict?)

WTFKr0:
Yh merge conflict with the date filter I think

PR #791, opened 2015-12-26
Analysis

### LINKED ISSUE #608: instant per-match automatic analysis/advice
An idea that might be cool:

After each match is parsed, write logic that can instantly analyze certain metrics based on hero played and performance.  For example, we can look at LH/EFF@10, then inform the player that their performance in that area was lacking for the hero they played.

Things we can look at:
* LH/EFF
* Kill drought (gap in kill times) for a ganking hero
* farming drought (low gold earned delta over an interval) for a farming hero
* Flaming in all chat
* lack of warding/min
* Excessive pinging
* Courier feeding
* low ability accuracy
* items not used but bought
* slow item timing (might have to be specific hero-item pairs)

Might involve having to hardcode some heroes as "supports", or we could use Valve's "support" designations, although those might be out of date.  Alternatively, we could use our data to compute averages for each hero and compare against that.  We'd probably want to compute that on an interval/startup and save to redis or something.

howardchung:
fixes #608 

Still looking for feedback on the UI design and/or ideas for more triggers.


jgrayson676:
For the `isSupport` check: would it also be worth comparing something like number of lane creep kills at 10 minutes to what teammates have? I would expect that in most games the support(s) would have the least amount of lane farm, especially before 10 minutes (or some similarly early point). There are situations where an offlaner might move to jungle early leaving the support to farm the offlane which could throw this off, but I think for a lot of cases this would be a decent estimation.

(if this belongs in the issue thread then I'd be happy to move it there)

howardchung:
That's something we could add if the current check returns too many false positives.

howardchung:
@albertcui wants a blog post explaining the new feature.

albertcui:
Blog/documentation. There needs to be a "How are these suggestions
determined?" or something linked on the analysis page.

howardchung:
Would we have to keep this blogpost updated if we add new checks/change the criteria?  I don't want to have to maintain both the code and the blogpost individually.

Issue #794, opened 2016-01-02
Where to find throttling options


schadowfox:
Where can I set throttling options?

As right now I just get data for 10-20 seconds and then Steam Returns a 503, probably because yasp hits the servers more than 1req/sec

5:28:53 PM scanner.1   |  2016-01-02T16:28:53.062Z - info: getData: http://api.steampowered.com/IDOTA2Match_570/GetMatchHistoryBySequenceNum/V001/?key=My_API_KEY_HERE&start_at_match_seq_num=1804646739
5:28:53 PM skill.1     |  total results: 0, added: 0

5:29:19 PM scanner.1   |  2016-01-02T16:29:19.376Z - info: invalid response, retrying: http://api.steampowered.com/IDOTA2Match_570/GetMatchHistoryBySequenceNum/V001/?key=My_API_KEY_HERE&start_at_match_seq_num=1804648804

howardchung:
You can check `utility.js`  getData method which has a built-in delay.

schadowfox:
Thanks, setting this to a higher value fixed the issue as skill and scanner both hit the api at the same time - I was looking in config.js but didn't find that there.

howardchung:
This could be made into a configurable value if you want to submit a PR for
it and refactor the delay to config.js.

PR #795, opened 2016-01-02
addresses #794

schadowfox:


scanner.js
```diff
@@ -85,7 +85,7 @@ function scanApi(seq_num) {
         userPlayers = result.userPlayers;
         getData({
             url: container.url,
-            delay: 1000
+            delay: config.DELAY_SCANNER
```
howardchung: Number() it?

utility.js
```diff
@@ -129,7 +129,7 @@ function generateJob(type, payload) {
 
 function getData(url, cb) {
     var u;
-    var delay = 1000;
+    var delay = config.DELAY_UTILITY;
```
howardchung: actually this will use a passed in delay if `url` is a object and has delay property.

So, rename the config to DEFAULT_DELAY, and get rid of the scanner one (since it's using this already).

schadowfox:
Changed it~

scanner.js
```diff
@@ -85,7 +85,7 @@ function scanApi(seq_num) {
         userPlayers = result.userPlayers;
         getData({
             url: container.url,
-            delay: 1000
+            delay: Number('config.DEFAULT_DELAY')
```
howardchung: no quotes

PR #802, opened 2016-01-04
Improved CSS for the footer

crisbal:
Hello, 

I decided to change a little bit the CSS for the footer of the website because, right now, I think it has one or two things that could be improved.

Current Footer 
====

**Desktop Version**

The only problem is about uneven padding. This results in the footer not getting centered.

A bit difficult to notice here but on the CSS you can see it. It was `padding: 1.3em 0 0 1em;` for no apparent reasons.

footerDesktopCurrent

**Mobile/Small Screen Version**

The footer on mobile is not centered (not only because of the padding) and I think that the lists are not the best looking.

footerMobileCurrent

My improvement 
====

**Desktop**

Even Padding and the footer is now centered for real.

Again: a bit hard to notice from a picture.

 footerDesktopMine

**Mobile**

I removed the bullet points from the lists and centered them.

 footerMobileMine

Let me know what you think about this. I tested it a bit and it should not break anything.

### DIFF
--- a/public/css/yasp.css
+++ b/public/css/yasp.css
@@ -28,8 +28,16 @@ footer {
     bottom: 0;
     background-color: #000;
     width: 100%;
-    padding: 1.3em 0 0 1em;
+    padding: 1.3em;
 }
+
+@media (max-width: 991px) {
+    footer ul{
+        text-align: center;
+        list-style-type: none;
+        padding: 0;
+    }
+} 
 
 h3 img {
     padding-right: 1em;


crisbal:
Also, probably off topic for this pull request, is there any way to run a local copy of yasp without the database and java/maven just to test front end (html/css) stuff?

howardchung:
Thanks!  @albertcui should review this.

Re: database, unfortunately not.  I think we'd have to set up web to be able to run with dummy data and not try to read/write from database.

albertcui:
Awesome looks good. Thanks!

PR #867, opened 2016-02-08
Completehistory

### LINKED ISSUE #359: Complete history
We could potentially add the entire Dota 2 match history.

Back of envelope math says:

UPDATED (2016-1-13):

Time:
~2 billion seqnums, probably ~700 million matches
An API request will get 100 at a time
If we make one request a second, it will take about 57 days to import.
We can parallelize by starting multiple importer threads (start at 0, 100 million, etc)

Storage:
If we remove the ability_upgrades data from each match:
Each player_match has about 25 fields, mostly 4 byte ints.  Each match has about another 20 fields, 4 byte ints, so ~1080 bytes/match.
700 million matches*1KB=700GB.

Scanner:
We would want to insert all matches, but only parse those with an active user (same as now).
There are ~1 million matches played per day (more on weekends and after patches).
Dropping the ability_upgrades data (store it in Redis with TTL):
Redis:
1 million * 1000 bytes per match * 7 days = 7GB.
DB:
1 million * 1kb =1GB of growth per day.
Parsing:
50k * 200KB = 10GB of growth per day.

11GB is sustainable.  It seems like the bulk of our insertion storage will come from parsing rather than the basic match data for all matches.

Possible impacts:
* Index growth.  Our indexes will likely get quite a bit bigger due to the 7x increase in number of rows in matches and player_matches.

If we do this, we'll probably need to resize our drive to 800GB or larger. (Currently 2TB).

- [x] enable add all matches
- [x] track all matches start seq num, 1876742099
- [x] run script to backfill matches
- [x] update FAQ
- [x] update player profile tooltips
- [x] ~~delete fullhistory from sign-in flow~~

howardchung:
fixes #359 

This PR updates the FAQ and player profiles.  Merge after complete history is done (expected to take 40-60 days)


howardchung:
Doing this separately (branches have diverged)

PR #877, opened 2016-02-14
Pros

### LINKED ISSUE #602: tracking notable players
- [x] How are we going to keep it maintained?  Professional players come and go all the time. (Steam API)
- [x] Do we want to fetch full history for these players?  If so, can we automate full history for new players on the list? (automatically done if we fetch complete history)
- [ ] Add a tab for pro players featured?

hi
The general idea is to have a list similar to the cheese list for professional players or highly valued people in the dota scene. The reasoning is that unlike for the normal average player there are usually a few hundred people who interested in their performance.

I got this idea recently when synderen was streaming and asked what his winrate with clockwerk versus earth spirit is. I figured that is easy to find out with yasp, while that is true it doesn't really works if yasp only knows of a small portion of his games.

The question can yasp afford that? Tracking and especially parsing is quite resource intensive.

### LINKED ISSUE #701: Bring back MMR leaderboard
Make leader board of top X (100?) active YASP users. This will be useful for users to see the parsed matches of high skilled players.

Add a histogram of MMR.

howardchung:
fixes #701 
fixes #602 

- [x] need to compute leaderboard in background.  It's too slow to run the query otherwise.


howardchung:
rolling into rankings PR

PR #885, opened 2016-02-17
Replace unicode char with fontawesome.io

### LINKED ISSUE #744: Death icon on "Teamfights" view doesn't render.
On all of my machines (two different physical machines running both windows and linux), using the latest version of Google Chrome (46.0.2490.80 as of the creation of this issue), the icon used for showing death in teamfights does not render properly. Instead of a skull, I am presented with an error box (shown below).

The cause of this issue is that Chrome, by default, does not support UTF-8 encoding -- you have to enable it in *Settings -> Advanced Settings -> Web Content -> Customize fonts*. It would be better to generate icons like this using SVG, so as to prevent the general user from experiencing the same.

missing_death_icon

bguzryanto:
fixes #744 

Fix for #744

### DIFF
--- a/views/match/match_teamfights.jade
+++ b/views/match/match_teamfights.jade
@@ -90,13 +90,13 @@ block match_content
               +hero_td(player)
               td
                 if player.participate
-                  div(style="font-size:30px;") !{"&#10003;"}
+                  i.fa.fa-check(style="font-size:28px;")
               td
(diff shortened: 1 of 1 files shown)


howardchung:
@albertcui review

albertcui:
Looks good. Thanks!

PR #891, opened 2016-02-21
Add message for players who abandoned the match.

### LINKED ISSUE #886: 'Abandoned' flagging
When showing a match, Dotabuff puts a little 'abandoned' note near a player's name if they abandoned it.

Is is possible to do something similar in Yasp?

JazzyJosh:
fixes #886 

Addresses yasp-dota/yasp#886

### DIFF
--- a/views/match/match_index.jade
+++ b/views/match/match_index.jade
@@ -38,7 +38,7 @@ block match_content
               th: abbr(title=tooltips.hero_healing) HH
               th: abbr(title=tooltips.items) Items
           tr(class=player.isRadiant ? "radiant" : "dire")
-            +hero_td(player)
+            +hero_td(player, player.abandons ? constants.leaver_status[player.leaver_status] : "")
             td
               //img.img-sm(src=player.avatar)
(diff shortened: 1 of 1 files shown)


JazzyJosh:
@howardchung I know you assigned this to yourself, but I wanted to get my environment running anyway. Feel free to decline if you've already fixed this.

howardchung:
I haven't, thanks for looking at this.

* Looks like you forgot to include the change to hero_td mixin.
* Since you're checking a property on player anyway, check it within hero_td instead of passing a param to hero_td.

JazzyJosh:
I didn't make any changes to the mixin. I reused it the same way that match_index was, unless I'm missing something.

I could refactor those usages to separate mixins or just allow hero_td to have a block following it. I assume the first option would be preferred.

howardchung:
ah I see you're reusing the metadata field.

I think it's fine to show up only on the index page.

Looks okay to me, @albertcui please review.

albertcui:
Looks good to me. Thanks!

PR #902, opened 2016-02-28
fixed profiler requesting empty steamids

schadowfox:
checking if steamids is not empty before starting to profile.

### DIFF
--- a/profiler.js
+++ b/profiler.js
@@ -42,6 +42,11 @@ function getSummaries(cb)
             {
                 return cb(err);
             }
+	    if (results.rows.length === 0)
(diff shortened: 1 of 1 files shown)


profiler.js
```diff
@@ -42,6 +42,11 @@ function getSummaries(cb)
             {
                 return cb(err);
             }
+	    if (results.rows.length === 0)
```
howardchung: I prefer !results.rows.length but this is okay too

howardchung:

albertcui:
Looks good. Thanks!

PR #905, opened 2016-02-29
only add players to ranker queue when ENABLE_RANKER is set in config

schadowfox:
on a small vps / lowmem instance the new ranker will crash redis (happened to me frequently), 
most people don't have alot of mmr data to begin with,
so the ranker shouldn't be enabled by default.

### DIFF
--- a/scanner.js
+++ b/scanner.js
@@ -148,7 +148,7 @@ function scanApi(seq_num)
                     {
                         "decideMmr": function(cb)
                         {
-                            if (match.lobby_type === 7 && p.account_id !== constants.anonymous_account_id && (p.account_id in userPlayers || (config.ENABLE_RANDOM_MMR_UPDATE && match.match_id % 10 === 0)))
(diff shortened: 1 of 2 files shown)


scanner.js
```diff
@@ -148,7 +148,7 @@ function scanApi(seq_num)
                     {
                         "decideMmr": function(cb)
                         {
-                            if (match.lobby_type === 7 && p.account_id !== constants.anonymous_account_id && (p.account_id in userPlayers || (config.ENABLE_RANDOM_MMR_UPDATE && match.match_id % 10 === 0)))
+                            if (match.lobby_type === 7 && p.account_id !== constants.anonymous_account_id && (p.account_id in userPlayers && config.ENABLE_RANKER || (config.ENABLE_RANDOM_MMR_UPDATE && match.match_id % 10 === 0 && config.ENABLE_RANKER)))
```
howardchung: you don't need to add ranker feature switch here.

scanner.js
```diff
@@ -148,7 +148,7 @@ function scanApi(seq_num)
                     {
```
howardchung: If you are adding a new feature switch also put it in config.js for consistency.

PR #907, opened 2016-03-02
only add players to ranker queue when ENABLE_RANKER is set in config

schadowfox:
on a small vps / lowmem instance the new ranker will crash redis (happened to me frequently), 
most people don't have alot of mmr data to begin with,
so the ranker shouldn't be enabled by default.


howardchung:
signoff.  @albertcui is on vacation until next thursday

albertcui:
Looks good to me! Thanks.

PR #929, opened 2016-03-22
add path to header

howardchung:
@albertcui's improvements.

Making the PR so we don't forget about this.


albertcui:
Oh, this is already on master. I think I commited/pushed to the wrong branch, then reset the commit, checked out master, and recommited/pushed.

PR #930, opened 2016-03-22
Change Item Active detection

WTFKr0:
Adds ring_of_aquila, ring_of_basilius and linken's shpere to Active items
Signed-off-by: WTFKr0 <thomas.kovatchitch@gmail.com>

### DIFF
--- a/analysis.js
+++ b/analysis.js
@@ -326,6 +326,6 @@ function generatePlayerAnalysis(match, pm)
             "bloodstone": 1,
             "radiance": 1
         };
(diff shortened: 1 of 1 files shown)


howardchung:
Do we expect players to use those items at least once?

WTFKr0:
Yeah, that's a good question !
Linken maybe ?
It's like Quelling Blade / Battle Fury / Mjollnir : don't need to be used systematically

howardchung:
@albertcui ?

I'm fine with it either way

albertcui:
Fine with me. Thanks!

PR #934, opened 2016-03-23
initial implementation of match percentiles

### LINKED ISSUE #923: Compute average MMR and percentile for matches
Wouldn't it be more interesting and/or helpful to have the players' skill level based on the other players like the IQ is calculated? Also, for this you should calculate the average(the exact integer, not the range) every week or so to keep it updated and store the mmr in a match, not the skill range, because the average may change and updating the skill for every match won't be a great thing.

The system should work like this: 
- store average mmr of the players in the match(10 players) as a match property(match.avg/skill/mmr)
- when someone sees that match in the matches/overview tab, compare the match average with the average of everyone(10000000 players)

Example of skill ranges in case the average is 3k:
    below average(100 to 35% below)
    low average(35 to 25% below)
    average(25% below and 25% above 3k) = 50% of the players
    high average(25 to 35% above)
    above average(35 to 100% average)

Something like this: https://qph.is.quoracdn.net/main-qimg-a5a6f1c8861ec17a6d5d3727aa1b693a?convert_to_webp=true

PS: I know that you only have data for the players that both have their mmr set as public and have an yasp account. I still think this may be a good idea.

EDIT:
This will work only for ranked matches. For normal matches you can either keep the current system or try to talk to Valve into giving you acces to the private mmr or use my system idea by making an average based on the other players in the match if they calibrated.

howardchung:
fixes #923

Schema changes:
* `alter table matches drop column url;`

Reviewers:
@albertcui 
@vali19n

### DIFF
--- a/benchmarkMatch.js
+++ b/benchmarkMatch.js
@@ -1,4 +1,5 @@
 var moment = require('moment');
+var utility = require('./utility');
 var config = require('./config');
 var benchmarks = require('./benchmarks');
 var async = require('async');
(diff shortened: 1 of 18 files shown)


howardchung:
@albertcui I don't know if you want to roll this into the April release or delay it.  If we want to delay it I think we're going to need to hide the UI element under feature switch.

PR #954, opened 2016-04-03
add player searching

### LINKED ISSUE #835: player search by name
I built a text index on the personaname field so we can search for players by name.  Is this something we want to implement?  Currently if you search for a random player they are unlikely to have all matches listed (unless they have signed in)

Example query:
```
select * from players where personaname ILIKE '%sniffle%' LIMIT 1000;
```

albertcui:
Should add tests, but think the style is good to go.

Fixes #835


public/css/search.css
```diff
@@ -0,0 +1,157 @@
+.grid {
```
howardchung: can we refactor this somewhere?  I think we're reusing it a lot.

queries.js
```diff
@@ -832,6 +832,49 @@ function mmrEstimate(db, redis, account_id, cb)
...
     });
 }
+
+function searchPlayer(db, query, cb)
+{
+    async.parallel(
+        {
+            account_id: function(callback)
+            {
+                if (isNaN(query))
```
howardchung: Number.isNaN is more robust.  Use that

queries.js
```diff
@@ -832,6 +832,49 @@ function mmrEstimate(db, redis, account_id, cb)
...
+                {
+                    return callback();
+                } else
```
howardchung: Braces on own line, if you want to adhere to styling guidelines.

web.js
```diff
@@ -415,6 +415,26 @@ app.get('/benchmarks/:hero_id?', function(req, res, cb)
...
+            {
+                cb(err);
+            }
+
+            return res.render('search', {
+                query: req.query.q,
+                result: result
+            })
+        })
+    } else {
```
howardchung: braces on own line

web.js
```diff
@@ -415,6 +415,26 @@ app.get('/benchmarks/:hero_id?', function(req, res, cb)
...
+        {
+            if (err)
+            {
+                cb(err);
+            }
+
+            return res.render('search', {
+                query: req.query.q,
+                result: result
+            })
```
howardchung: semicolon

queries.js
```diff
@@ -912,6 +912,95 @@ function mmrEstimate(db, redis, account_id, cb)
...
+            },
+            steam64: function(callback)
+            {
+                if (!query.startsWith("7656119"))
+                {
+                    return callback();
+                }
+                else
+                {
+                    findPlayer(db, {steamid: query}, callback);
```
howardchung: Don't do this.  Convert the 64 bit steam ID to 32 bit account ID and search via that.

We don't have an index on steamid.

queries.js
```diff
@@ -912,6 +912,95 @@ function mmrEstimate(db, redis, account_id, cb)
...
+                }
+            },
+            url: function(callback)
+            {
+                if (query.indexOf("http://steamcommunity.com") === -1) {
+                    return callback();
+                }
+                else
+                {
+                    findPlayer(db, {profileurl: query}, callback);
```
howardchung: this won't work unless players search for entire URLs?  That usually won't be the case?

PR #957, opened 2016-04-06
remove april from nav, places until /april route.

albertcui:


web.js
```diff
@@ -415,13 +414,14 @@ app.get('/benchmarks/:hero_id?', function(req, res, cb)
         });
     }
 });
-app.get('/plusplus', function(req, res, cb)
+app.get('/april/:year?', function(req, res, cb)
```
howardchung: Are we taking this with us to react?

howardchung:
signoff

web.js
albertcui: Not sure yet. Might just screenshot.

PR #984, opened 2016-04-23
Topmatches

### LINKED ISSUE #951: Improve code organization and style
- [x] improve code organization - we have too many files, and too many top level files.
- [x] we should probably establish/adhere to an agreed code and code organization style.

### LINKED ISSUE #991: Search fails on short queries
Search doesn't work when the input is very short. It won't work at all for me for two character queries, and only some three character queries work.

Does a GIN index exist on the players.personaname column?

Assuming I'm looking at the correct piece of code ([from here](https://github.com/yasp-dota/yasp/blob/master/queries.js#L946)):

```
                db.raw(`
                    SELECT account_id, personaname, avatarfull, similarity(personaname, ?)
                    FROM players WHERE personaname ILIKE ?
                    ORDER BY similarity DESC LIMIT 100
                    `, [query, "%" + query + "%"])
```

Have you ensured that including ILIKE in the query gives performance improvements compared to leaving it out, or compared to using the [similarity operator](http://www.postgresql.org/docs/9.1/static/pgtrgm.html) (%)? If you are pre Postgres 9.1 then ILIKE will be a significant bottleneck in this case.

### LINKED ISSUE #998: Fix off by one error in MMR distribution
IIRC saw this on Reddit.  We currently bucket users down to the nearest 100, but the percentile considers users equal or less than the next 100 up.

### LINKED ISSUE #1002: Laning map shouldn't count position on paused time
When the match is paused early on for a long time, the laning map shows the position where each hero was on paused time as the main position on heatmap. It should ignore the paused time position.

Example:
https://yasp.co/matches/2324675156/performances

In the example, the earthshaker has normal heatmap because he entered the game after the pause.

howardchung:
fixes #951 (organization)
fixes #998 (OBO error in mmr distribution)
fixes #991 (short search queries)
fixes #1002 (duplicate entity interval data)

* Ability to sort/limit matches in filter.
* Decommission fullhistory
* Decommission some SQL-based queries for distributions
* Setup for full Cassandra (if we want to try it)
* Speed up status page via more Redis

Things to watch out for:
* backend node.  Are services restart looping?
* Parser counts.  Make sure number of nodes stays steady


albertcui:
Signoff, but has merge conflicts ;p

Issue #994, opened 2016-04-25
Broken hero selector in safari


howardchung:
Images are set to 100% window height for some reason.

PR #1003, opened 2016-04-29
#994 - Removed min-height from hero images on the benchmark page

coreymaher:
Should fix the issue in #994 
Before:
issue 994 before

After:
issue 994 after


howardchung:
@albertcui review

albertcui:
Great! Thanks so much :dancer:

Issue #875, opened 2016-02-14
Sort order of skill is incorrect.


aoisensi:
![](https://i.gyazo.com/bbb729278de0756466ab9b78b4b75d0e.png)

howardchung:
I think it's just alphabetical.  Could define custom sort order, I suppose.

likered:
A normal player will probably be in one skill bracket for the majority of the time, only at a certain time period will the player actually move from one skill bracket to the next (for example, Normal -> High Skill or High Skill -> Very High Skill). So I don't think this will be very applicable most of the time. Of course, you could argue that if you play in a party with a friend(s) who's very high or low in skill will change your skill bracket.

howardchung:
Personally I see quite a bit of fluctuation myself but considering there
are only 4 possible values sorting isn't a huge deal

aoisensi:
:+1:

PR #1004, opened 2016-04-29
#875 - Correctly sort match skill in match tables

coreymaher:
Fixes the sort order of the skill column in the matches table as described in issue #875.

I don't really like hard coding the skills again, but I couldn't figure out how to make the skill constants available in views/player/player.jade

### DIFF
--- a/views/player/player.jade
+++ b/views/player/player.jade
@@ -88,6 +88,22 @@ append footer_assets
         },
         columnDefs: [
             {
+                targets: "skill",
+                render: function(data, type) {
+                    var skillIDs = {
+                        1: "Normal",
+                        2: "High",
+                        3: "Very High",
+                    };
+
+                    if (type === "display") {
+                        return (data in skillIDs ? skillIDs[data] : data);
+                    }
+
+                    return data;
+                }
+            },
+            {
                 "targets": "fromNow",
                 render: function(data, type) {
                     if (type === "display") {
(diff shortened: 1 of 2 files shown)


howardchung:
OK, but this will break if we ever add internationalization.  A better approach might be to render the number server side and then trnwslate that to a string on the client side.

coreymaher:
Good point. I'll make that change.

coreymaher:
@howardchung I've made the change you suggested. Let me know what you think.

views/player/player.jade
```diff
@@ -88,6 +88,22 @@ append footer_assets
...
+                targets: "skill",
+                render: function(data, type) {
+                    var skillIDs = {
+                        1: "Normal",
+                        2: "High",
+                        3: "Very High",
+                    };
+
+                    if (type === "display") {
+                        return (data in skillIDs ? skillIDs[data] : '');
```
howardchung: Default to data instead if empty string.

coreymaher:
Alright it now defaults to data.

howardchung:
signoff

albertcui:
Thanks!

howardchung:
@coreymaher for future reference if you start your OP with something like:

fixes #875 

it will auto close the issue when merged.

coreymaher:
Looks like the last merge didn't work quite right. This pull request should fix it: #1011

PR #1008, opened 2016-04-30
#873 - Added totals to the "Damage Done/Taken" table on the match combat tab

### LINKED ISSUE #873: Sum damage crosstables
There should be a sum total at the bottom/right side where the total Damage Taken/Done per hero is shown. What I mean is in for example: https://yasp.co/matches/2147794398/combat

Sand King's total damage taken/done to all heroes should be shown on a new column on the far right as 15/23.3 and Huskar's total damage taken/done  should be shown in the first column in a bottom row as 26.9/24.4

coreymaher:
fixes #873

### DIFF
--- a/util/compute.js
+++ b/util/compute.js
@@ -396,6 +396,17 @@ function count_words(player_match, player_filter)
 function renderMatch(m)
 {
     //do render-only processing (not needed for aggregation, only for match display)
+    m.hero_combat = {
+        damage: {
+            radiant: 0,
+            dire: 0,
+        },
+        kills: {
+            radiant: 0,
+            dire: 0,
+        },
+    };
+
     m.players.forEach(function(player_match, i)
     {
         //converts hashes to arrays and sorts them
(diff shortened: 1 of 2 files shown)


views/match/match_combat.jade
```diff
@@ -35,32 +35,64 @@ block match_content
     .col-md-6
```
howardchung: You may as well do this for the Kills crosstable as well.

coreymaher:
Sure that should be easier

views/match/match_combat.jade
```diff
@@ -35,32 +35,64 @@ block match_content
...
-        table.table.table-hover.table-condensed
+        table.table.table-hover.table-condensed.combat_damage
           thead
             tr
               th: abbr(title=tooltips.hero_id) Hero
               each player, ind in match.players.slice(match.players.length/2, match.players.length)
                 +hero_td(player)
+              th Totals
           tbody
+            - var radiant_total_damage = 0;
```
howardchung: It might be cleaner to do this computation in `compute.js`.  Jade isn't really suited to heavy-duty code since it gets hard to read.

coreymaher:
I've added totals to the kill/death table and refactored a lot of the jade logic into compute.js
Let me know what you think.

howardchung:
signoff

albertcui:
Great, thanks!

util/compute.js
```diff
@@ -475,6 +486,53 @@ function renderMatch(m)
...
+        m.players.forEach(function(other_pm) {
+            var team = (player_match.isRadiant) ? 'radiant' : 'dire';
+            var other_hero = constants.heroes[other_pm.hero_id];
+            var damage = 0;
+            var taken  = 0;
+            var kills  = 0;
+            var deaths = 0;
+
+            // Only care about enemy hero combat
+            if (player_match.isRadiant !== other_pm.isRadiant) {
```
howardchung: exception occurring if !player_match.damage

howardchung:
exception causing crashes.  I think I fixed the issue, building now.

howardchung:
In the future we may want to consider:
* wrapping code in a try/catch block so the webserver doesn't crash if there's a bug
* making sure there's a test to cover the scenario so CI catches this

Issue #801, opened 2016-01-03
Shadow razes display the same one


Jemsuper:
Even though the site is able to differentiate between all 3 razes, it only shows the close one for all 3. 
(in skills used) https://yasp.co/matches/2049875930/performances

howardchung:
Need to manually patch the tooltip.  Data is from http://www.dota2.com/jsfeed/abilitydata

Derek-Li:
Under constants.js
    abilities.nevermore_shadowraze2 = abilities.nevermore_shadowraze1;
    abilities.nevermore_shadowraze3 = abilities.nevermore_shadowraze1;

I'm guessing this the spot where it causes Shadow Raze 2/3 to be shown as Shadow Raze 1?

howardchung:
Yes.  The tooltip data does not contain separate entries for the three razes so we wuld need to make copies of shadowraze1, edit them, and point to them.

PR #1009, opened 2016-04-30
#801 - Added SF shadowraze2 and shadowraze3 abilities

coreymaher:
Fixes the issue described in #801 

I assume json/abilities.json is grabbed from Valve. If so, we'll need to make sure we keep these updates the next time the abilities file is regenerated, or we'll need a way to add/augment the these files.
Let me know what you think.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -72,13 +72,24 @@ countries.forEach(function(c){
 constants.items = items;
 //abilities, already keyed by name
 var abilities = constants.abilities.abilitydata;
+
+// Add missing Shadow Fiend raze abilities by copying the shortest raze
+if (!abilities.nevermore_shadowraze2) {
+    abilities.nevermore_shadowraze2 = Object.assign({}, abilities.nevermore_shadowraze1);
+    // Find and replace short raze range with medium raze range
+    abilities.nevermore_shadowraze2.attrib = abilities.nevermore_shadowraze2.attrib.replace(/\d{3}/, 450);
+}
+if (!abilities.nevermore_shadowraze3) {
+    abilities.nevermore_shadowraze3 = Object.assign({}, abilities.nevermore_shadowraze1);
+    // Find and replace short raze range with long raze range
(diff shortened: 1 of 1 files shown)


constants.js
```diff
@@ -77,8 +77,6 @@ for (var key2 in abilities)
     abilities[key2].img = "/apps/dota2/images/abilities/" + key2 + "_md.png";
     abilities[key2].cmb = abilities[key2].cmb.replace("http://cdn.dota2.com", "");
 }
```
howardchung: This script grabs the constants from JSON and applies the additional transforms to it.  You'll probably want to patch the text here instead of directly in the JSON file (since that will get overwritten by updateconstants).

Also to pick up changes made to shadowraze1 you'll may want to edit the data based on the value of shadowraze1.

constants.js
```diff
@@ -72,13 +72,22 @@ countries.forEach(function(c){
 constants.items = items;
 //abilities, already keyed by name
 var abilities = constants.abilities.abilitydata;
+
+// Add missing Shadow Fiend raze abilities by copying the shortest raze
+if (!abilities.nevermore_shadowraze2) {
+    abilities.nevermore_shadowraze2 = Object.assign({}, abilities.nevermore_shadowraze1);
+    abilities.nevermore_shadowraze2.attrib = abilities.nevermore_shadowraze2.attrib.replace(/\d{3}/, 450);
```
howardchung: you may want to comment this bit of magic regex for readability

coreymaher:
Sure thing. I've added some comments to make it more clear what those lines are doing. I've also rebased onto master.

Let me know what you think.

howardchung:
thumbs up = signoff 
@albertcui should review

albertcui:
Thanks!

howardchung:
I don't think this made it into the (hotfix) build I'm trying to deploy.  Should be autodeployed in 24 hours.

PR #1018, opened 2016-05-04
Stream api

### LINKED ISSUE #1013: replay url API endpoint
As a feature, it'd be cool if there was a way to make a HTTP GET request to http://api.yasp.co/matches/$matchid/parse/stream that returns a series of JSON-encoded Entry object, (potentially terminated by a {"state":"finished"} blob).

Each time this endpoint is hit, it should fire a reparse if the parser has changed (or if it's a cache-miss).

(just for clarity-sake, the GET request from the client should probably have an overridden higher timeout time just in case there is a > 30 second wait for a data stream).

TODO
- [x] finalize API
- [ ] enable support for S1 replays (need sample replays to test against, missing entity data)  https://github.com/yasp-dota/yasp/blob/23bd8a8a36814fe771a57d2eba770d4b14007b0d/java_parser/src/main/java/yasp/Main.java

howardchung:
fixes #1013 

Reviewers:
@albertcui 
@noxville


Noxville:
Looks great ^

routes/api.js
```diff
@@ -288,5 +292,47 @@ module.exports = function(db, redis, cassandra)
...
+        }, function(err)
+        {
+            if (err)
+            {
+                return cb(err);
+            }
+            res.setHeader('Content-Type', 'text/event-stream');
+            res.setHeader('Cache-Control', 'no-cache');
+            subscriber.subscribe('logParse:' + identifier + ':' + req.params.match_id);
+            subscriber.on('message', function(channel, message)
```
albertcui: You need to check the channel.

routes/api.js
```diff
@@ -288,5 +292,47 @@ module.exports = function(db, redis, cassandra)
...
+            {
+                if (message === "END")
+                {
+                    subscriber.unsubscribe('logParse:' + identifier + ':' + req.params.match_id);
+                    subscriber.quit();
```
howardchung: Don't quit() since reusing connection

PR #1023, opened 2016-05-06
No storedeps

howardchung:
Node caches require() calls, so we actually shouldn't need to do the thing we're doing where we pass references to db, redis, and cassandra around.


howardchung:
scratch this.  too much risk of regression

PR #1040, opened 2016-05-15
damage_inflictor_received

### LINKED ISSUE #903: Request parse of damage received.
A useful feature is the biggest hit feature, but the team and I often try to imagine the sources of the damage our hero has received and where the biggest source of damage is coming from. So perhaps we could have a feature that shows the amount of damage taken and have it split between what damaged us.

Eg. If a bristleback is focused - it would be nice to see how much damage he was able to soak up for the team.

howardchung:
fixes #903 

## Deployment
- [x] update cassandra schema `ALTER TABLE player_matches add damage_inflictor_received text;`

Reviewers:
@albertcui 
@mm11wils

Scheduled release:
2016-06-03?


don4of4:
Just a friendly tip:  it might be worth reviewing and making use of interactive rebasing for this project, it would help keep the commit log cleaner.

https://www.atlassian.com/git/tutorials/rewriting-history/

howardchung:
Thanks for the tip.  I agree it's kind of messy

howardchung:
Tried to rebase, hopefully I did it right :)

don4of4:
Generally an interactive rebase requires a force push and should be done on a local/user branch.  Now, I do not see any changes here, so I don't think you managed to push anything.

howardchung:
It went from 14 commits down to 9?  I suppose it could be squashed down to 1.

(I did do a force push)

don4of4:
Oh... You should def. squash commits which are called "undo" etc.

PR #1045, opened 2016-05-20
fixes #300

### LINKED ISSUE #300: parallelize sequential api
This might help us avoid falling behind so much when the Steam API slows down.  I don't know if it's necessary, but I'm outlining an implementation here.

OLD design:
* ~~Code that is polling the sequential API should be refactored into a separate worker file (scanner.js) which can be distributed.~~ Done
* ~~When a match to add is detected, we queue a new job type called "insert", which YASP core's worker.js is processing.~~
* When we make a API request, we save in redis the sequence number we started at.  We can assume that the next 100 numbers will be processed.  What if we're at the leading edge and there are <100 matches though?  We could keep retrying the same seq_num until we get 100 matches.
* On each sequential api request, we take the current progress in redis and add 100 (the minimum number of sequence numbers so we are guaranteed not to miss any matches) to get the number we want to start at, then make the API request.  Once a request is made, we need to update redis to tell subsequent requests to start at a higher number.  We may need two vars, one to track what matches we've actually added, and one to track the number we should start at.
* ~~The API scan code currently uses local variables trackedPlayers and ratingPlayers to determine matches we need to request MMR/get replay on.  We would need to access redis for these variables to allow shared state between multiple hosts.~~ (fixed)
* It is possible for us to get matches in duplicate (and quite likely).  In this case, since we only have a single thread inserting, we should detect that we already have a replay url and not re-retrieve it.  We will still insert/update the db data, however.

NEW design:
- [x] start at match_seq_num, make multiple calls to steam API at sequence number n, n+100, n+200...
- [x] duplicates already handled  by storing recently seen match_ids in redis
- [x] if less than 100 matches, redo the page
- [x] for any single call, the next_seq_num should be current num + num_workers*100 + 1
- [ ] the trailing worker (ID 0) should update match_seq_num in redis

howardchung:
Steam API responses taking 15+ seconds lately.  This attempts to brute-force sequence numbers (100 per page), which allows us to make multiple requests in parallel.

The previous design required us to wait for a single call to return, then issued the next API call based on the last sequence number on the page.  This is the most economical design, but is extremely sensitive to slow API.

Reviewers:
@albertcui 
@nicholashh

### DIFF
--- a/config.js
+++ b/config.js
@@ -22,7 +22,7 @@ var defaults = {
     "SESSION_SECRET": "secret to encrypt cookies with", //string to encrypt cookies
     "ROOT_URL": "http://localhost:5000", //base url to redirect to after steam oauth login
     "WORK_URL": "http://localhost:5400", //url to request work from (for worker nodes)
-    "START_SEQ_NUM": "", //REDIS: use redis number, truthy: use sequence number, else: use auto
+    "START_SEQ_NUM": "", //truthy: use sequence number stored in redis, else: use approximate value from live API
     "NODE_ENV": "development",
     "FRONTEND_PORT": "5000",
     "RETRIEVER_PORT": "5100",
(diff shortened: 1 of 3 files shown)


howardchung:
LSI mitigation yolo

PR #1058, opened 2016-05-26
ReactSPA architecture/standards update

mike-robertson:
This is an example of what I think the ReactSPA should look like. Under public/javascript, the directories /components, /actions, /reducers are the major changes. I haven't updated every component to be in these new standards, and this is by no means necessarily an end product, but it's a look at how it could look. I think this structure will be much more maintainable/reusable and friendly to new contributors.

Additional changes that aren't here/I haven't had time for:

ESlint file: will drastically increase readability if followed.

Adding CSS-modules: In my opinion, this is the best solution to using css with react/components while still using true css.

Rest of the components: I didn't have time to do more than the Header.

Unit Tests: Don't want to waste time setting all this up if people done want these changes.

Probably other stuff.


public/js/actions/metadataActions.js
```diff
@@ -0,0 +1,34 @@
...
+const OK = 'yasp/metadata/OK';
+const ERROR = 'yasp/metadata/ERROR';
+
+export const metadataActions = {
+  REQUEST,
+  OK,
+  ERROR
+};
+
+const getMetadataRequest = () => ({ type: REQUST });
```
gu3st: This is a typo. Type should be REQUEST.

public/js/actions/userActions.js
```diff
@@ -0,0 +1,34 @@
...
+const OK = 'yasp/user/OK';
+const ERROR = 'yasp/user/ERROR';
+
+export const userActions = {
+  REQUEST,
+  OK,
+  ERROR
+};
+
+const getUserRequest = () => ({ type: REQUST });
```
gu3st: Same typo as with metadata actions

public/js/actions/index.js
```diff
@@ -0,0 +1,4 @@
+import { metadataActions, getMetadata } from './metadataActions';
```
howardchung: is it really necessary to have separate files for all the actions?  I kind of like the simplicity we have now.

public/js/components/Logo/Logo.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default () => {
+  return (
+    <div className="navbar-header">
+      <a href="/" className="navbar-brand">
+         <strong className="theme-blue">YASP</strong>
+         <a data-toggle="collapse" data-target=".navbar-collapse" className="navbar-toggle">
```
howardchung: React complains about the <a> in <a> nesting.  Maybe there's a better way to do this?

public/js/reducers/reducers.js
```diff
@@ -1,44 +1,32 @@
...
-      return Object.assign(
-      {}, state,
-      {
-        metadata:
-        {
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case actions.METADATA:
+      return Object.assign({}, state, {
```
howardchung: if we can use the spread operator, convert object.assign calls to use it

public/js/yaspv2.js
```diff
@@ -39,19 +41,20 @@ require('../../node_modules/dota2-minimap-hero-sprites/assets/stylesheets/dota2m
 //require('../../node_modules/bootswatch/darkly/bootstrap.css');
 //require('../css/yasp.css');
 const loggerMiddleware = createLogger();
-var reducer = combineReducers(Object.assign(
+const reducer = combineReducers(Object.assign(
 {},
 {
-  reducers: reducers,
+  [REDUCER_KEY]: appReducer,
```
howardchung: what is the purpose of having REDUCER_KEY?

webpack.config.js
```diff
@@ -15,12 +18,16 @@ module.exports = {
...
+      extensions: ['', '.jsx', '.js', '.css', '.json']
+    },
+    devtool: 'source-map',
     module:
     {
         loaders: [
             {
                 test: /\.css$/,
-                loader: "style-loader!css-loader"
+                loader: "style-loader!css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]d!postcss-loader"
```
howardchung: will this break the existing css loading?

webpack.config.js
```diff
@@ -29,30 +36,39 @@ module.exports = {
...
                 test: /\.(json)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
                 loader: "json-loader"
-                },
+            },
+            {
+                test: /\.(js|jsx)$/,
+                exclude: /(node_modules|bower_components)/,
+                loader: 'babel', // 'babel-loader' is also a legal name to reference
+                // include: path.resolve(process.cwd(), 'public/js'),
+                // query:
```
howardchung: why is this commented out?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
howardchung: is this file necessary?

public/js/components/Logo/Logo.jsx
mike-robertson: Agreed, I had left it as it was but I can change that to something else.

public/js/yaspv2.js
mike-robertson: It's mostly preference and good for cases where we would be exporting the reducer to use somewhere else (though I can't think of a use case for this spa to do so). Since we need to reference the reducer in our containers, I find it's easier to have a constant instead of remember it is under the reducer or something else which means changes to that property mean refactoring all the containers. But it's not necessary and pretty much preference.

webpack.config.js
mike-robertson: I'm not sure... maybe it's better to leave it out for now. I can't get css working locally even from a fresh pull (I'm not running the services or anything, just npm run webpack-dev-server), so I can't test. I'll remove this.

webpack.config.js
mike-robertson: moved to .babelrc since linting will also need that file (no need to have it in both places)

public/js/actions/index.js
mike-robertson: Yea, I was thinking the same thing because the factories are nice for these actions we have. Mostly I am used to this pattern/it felt easier to understand at a glance, and I sort of thought it might be nice to transform some of the data (camelCasing vs snake_case). Also since I haven't gotten services working locally yet I don't know exactly what all the data looks like (feels better to bring all that data to the top in actions rather than accessing deep properties). 

I'm not sure yet whether that outweighs the simplicity of the. We will definitely need lots more actions though as we add more stuff to the ui (pretty much all user interactions will trigger actions to update global store).

public/js/reducers/reducers.js
mike-robertson: I was planning on removing this file in favor of the 2 reducer files I have called gotMetadata and gotUser which use spread.

public/js/actions/index.js
howardchung: We are very limited in user actions though.  The data flow is 95% from server to user.

The only user inputs are:
* Match requests
* donations

public/js/yaspv2.js
howardchung: I think it's simpler to go with a single reducer and just have things under the 'reducer' key?  Would that also save the need to pass the name to the containers (they can all reference the same reducer?)

I think we don't have enough distinct reduce actions that we need to break it up.  The total is:
* 1 for match
* 1 for metadata
* ~10 for players
* maybe a few more for picks, rankings, benchmarks, etc.

howardchung:
@mike-robertson could you redo this against:
https://github.com/yasp-dota/ui

Issue #651, opened 2015-09-15
Friends matches


evaldobratti:
I've started to develop a functionality to find matches of a player and it's friends, and helps to track stacks, collect matches of you and your friends, etc. I have already done some job on my own, but doing everything alone it's too hard :(

A sample of this is deployed on http://45.55.183.239/friends/88738111,89224557

where on match page you can track matches of players in it by clicking on it's row
http://45.55.183.239/matches/1794495797 and go to the page above as well.

I'm wondering if you would accept a pull request of this functionality, then I'd would start to work on yasp

howardchung:
We actually kind of support this functionality already with the included account id filter.

evaldobratti:
Yeah, but I couldn't find a way to get matches that were played with another person, only the winrate and some other statistics

howardchung:
player pages are just views over a given query.  So if you visited the matches tab and then set the query you would get back all the matches that fit.

evaldobratti:
Ok, now I found it :P

The only thing that is left is the stats of the players in the query, but it the functionality is already done, as you previously sad.

Sorry boring you for this

howardchung:
Yeah, we're kind of focused on individual players, so if you wanted the stats of those players you'd have to visit their profiles.  Each profile is designed to be "centric" around that player.

evaldobratti:
Yeah, I got that
What I want to implement is a easier way to discover people that I have already played with, just get a match and see if I have already played with a person of that match (yeah, I want to blame them :P )

I got the yasp code yesterday at home, I'll start to develop something and if it fits/work well on yasp, may I open a pull request, ok?

howardchung:
Sure, feel free to open a PR.

I've thought about this problem already, though.  The problem is that getting the players that a certain player has played with is a relatively expensive operation.  Doing it for 10 players when we try to load a match page is probably going to be pretty slow.  Then after that we have to pairwise-check every pair on the team to see how often they play with each other, and "guess" the stack based on that.  

I suppose it MIGHT be possible to do this just by checking the players in the match against the currently logged in player.  We currently store a cache of "played with" players for users who have visited their profile.  However we may be moving to a cache of raw matches rather than a cache of aggregations so I don't know if we'll be able to support that in the future.

evaldobratti:
[...]

PR #1061, opened 2016-05-27
Played with

evaldobratti:
This is a WIP.
I have opened a request ages ago #651 about a easier way to find who we already played with.
I have already developed my own website for hobby (http://dotaparty.me/), but since I'm not going to pay more that I already pay for hosting services, I tryed to develop this functionality on yasp.
I have created a new tab on details match that show directly whom players have already played with.
It does not need any data migration, it's just a raw sql (so, there is one extra query when loading matches) .
I newbie on node and all its environment, so any feedback is appreciated.
It is still lacking tests, but I'll study and create them if this fuctionality is of your interest.

### DIFF
--- a/json/match_pages.json
+++ b/json/match_pages.json
@@ -61,5 +61,10 @@
   {
     "name": "Chat",
     "parsed": true
+  },
+  "mates":
+  {
+    "name": "Mates",
+    "new-feature": true
   }
 }
--- a/json/match_pages.json
+++ b/json/match_pages.json
@@ -62,9 +62,9 @@
     "name": "Chat",
     "parsed": true
   },
-  "mates":
+  "playedwith":
   {
-    "name": "Mates",
+    "name": "Played with",
     "new-feature": true
   }
 }
--- a/store/queries.js
+++ b/store/queries.js
@@ -506,40 +506,77 @@ function getMatch(db, redis, match_id, options, cb)
                         return cb(err);
                     }
                     match.players = players;
-                    computeMatchData(match);
-                    renderMatch(match);
-                    getMatchRating(redis, match, function(err, avg)
-                    {
+                    computeMates(match, function(err){
                         if (err)
                         {
                             return cb(err);
                         }
-                        var key = 'match_ratings:' + utility.getStartOfBlockHours(config.MATCH_RATING_RETENTION_HOURS, config.NODE_ENV === "development" ? 0 : -1);
-                        redis.zcard(key, function(err, card)
+                        computeMatchData(match);
+                        renderMatch(match);
+                        getMatchRating(redis, match, function(err, avg)
                         {
                             if (err)
                             {
                                 return cb(err);
                             }
-                            redis.zcount(key, 0, avg, function(err, count)
+                            var key = 'match_ratings:' + utility.getStartOfBlockHours(config.MATCH_RATING_RETENTION_HOURS, config.NODE_ENV === "development" ? 0 : -1);
+                            redis.zcard(key, function(err, card)
                             {
                                 if (err)
                                 {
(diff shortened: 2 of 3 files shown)


howardchung:
Considerations:

* I don't think this approach will scale.  Currently it looks like you're querying for all matches of about 10 players (twice), then counting the number of unique pairs encountered.  Based on the fact that it takes roughly 10-30 seconds to query for just one player's matches, this query will take minutes to run on production.
* We are moving data out of postgres and into cassandra, so this table won't exist anymore in a week or so.

* Potential alternate solution: Perhaps you can utilize the peers data for each player (https://yasp.co/api/players/88367253/peers).  That's still making 10 requests per match, though.

evaldobratti:
So, this feature could be added to yasp?

I've found some cassandra references in the code, but I thought the schema would almost be the same, I'll review the code to make it work on cassandra.
About the peers code, I studied it and maybe I have missed something, It seemed that all player matches was load and then the aggregation was make, is my understanding right?

If I use the api, it would be better to load data via ajax on client and fill the table as it loads, or render all the page on server?
Thank you :)

howardchung:
Cassandra does not support joins so you might not be able to do it the same way.

The peers page is now served out of a third table called "player_caches".  It is much faster to load a player's matches from this collection.

We are currently in the process of developing the new UI (see https://github.com/yasp-dota/ui), which is React-based and will use API endpoints on the server for everything.  If you want to implement this you should make it return as an endpoint and then work in that repo to build the UI for it.

Depending on how slow the query is, you may need to separate it from the match endpoint since that must load quickly for a good user experience.

evaldobratti:
I see, I'll checkout this new UI and problably, make it only available on it. I'll problably make a new endpoint and load it dynamically por each player, cus, as you said, it'll take about 10 queries to fill all the table.

Running yasp locally it's not fetching players data (as personaname and avatar), what could be missing?

howardchung:
Player data is not in matches.  It needs to be obtained from a separate endpoint (see `svc/profiler.js`).

Or you can manually fake some data

evaldobratti:
I'll check it later.
It's throwing 
```
...
 [299] ./~/font-awesome/css/font-awesome.css 1.21 kB {0} [built] [1 error]
...

ERROR in ./~/css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]d!./~/postcss-loader!./~/font-awesome/css/font-awesome.css
Module build failed: ReferenceError: Promise is not defined
    at LazyResult.async (/home/evaldo/codes/ui/node_modules/postcss-loader/node_modules/postcss/lib/lazy-result.js:157:31)
    at LazyResult.then (/home/evaldo/codes/ui/node_modules/postcss-loader/node_modules/postcss/lib/lazy-result.js:79:21)
    at Object.module.exports (/home/evaldo/codes/ui/node_modules/postcss-loader/index.js:63:10)
 @ ./~/font-awesome/css/font-awesome.css 4:14-179
```
when starting the new ui, any clues?
You can close this PR then

howardchung:
Check your node version and upgrade, most likely

evaldobratti:
Ok, thank you for your support :+1:

gu3st:
I think you need to upgrade your node version to at least v5, preferably v6 (as the UI, we are using pretty bleeding edge stuff like the spread operator)

evaldobratti:
I tried with node 6.2.0 and it did not work. On node 5.11.1 it seems ok. It is just basic html, with no css by now, right?

howardchung:
Try pulling latest master, I fixed a bug with it not working on 6.2.0 a few hours ago.

No CSS yet, we just started a few days ago

PR #1068, opened 2016-06-04
Minus stun time fix

Marvizm:


howardchung:
I'd prefer to just show the actual data.

Marvizm:
It's not first time when I saw an issue with negative stun value, I can't fix it in replay, but I can fix it here.

howardchung:
Yeah, but then we're hiding the actual data, even if we suspect it's wrong.

Marvizm:
You can stun on 4 sec or on 0 sec but not on -0.93 in dota 2. No need to hide data when it's easy to fix

howardchung:
My point is that we're obscuring the actual value in the replay, even if it's wrong.

Marvizm:
but it is stupid, if there is an error, it needs to be fixed

howardchung:
I don't know if it's an error.  I suspect it's a replay bug, but I have no way of knowing for sure.

@albertcui, @nicholashh ?

views/match/match_performances.jade
```diff
@@ -23,7 +23,7 @@ block match_content
             //td.activate(data-index=i, data-type="sen") #{player.sen_log.length || "-"}
             td.rankable #{(Number(player.lane_efficiency)*100).toFixed(2)}
             td.rankable #{player.lh_t && player.lh_t[10] ? player.lh_t[10] : "-"}
-            td.rankable #{Number(player.stuns).toFixed(2)}
+            td.rankable #{Number(player.stuns > 0 ? player.stuns : player.stuns == 0 ? 0 : 0).toFixed(2)}
```
howardchung: Should probably just Math.max(player.stuns, 0)

howardchung:
I suppose this is ok beause those interested in raw data can use the API.

@albertcui review

PR #1074, opened 2016-06-07
Added denies@10

### LINKED ISSUE #1049: Denies @10min
It would be nice to see denies at 10min in the performance tab. It would give a more detailed view who won the lane.

lomash:
fixes #1049 

I added the feature Denies @ 10 as mentioned in issue #1049. I'm not sure if I've covered everything but it seems to work


howardchung:
Awesome, this is the first time someone besides me has implemented a new parse feature.

We are storing the deny count every minute, which is more expensive than it strictly needs to be.  However it might be ok

lomash:
Should I just create just one instance at 10 instead?

howardchung:
Probably not needed.

howardchung:
@albertcui 

For bonus points, add an assert for this in the test

howardchung:
need conflict resolution @lomash 

also I'd like @albertcui to review and sign off on this

albertcui:
Signoff. Thanks!

howardchung:
The schema change needs to be run on production cluster prior to merge.
Maybe you or nick should practice that.

howardchung:
Note that there are also a few more changes to be made if you want to add a graph or add it to player page aggregations.

lomash:
Graphs for denies will not be that helpful in my opinion since they rarely happen after the laning phase

howardchung:
cqlsh command:
`ALTER TABLE player_matches ADD dn_t text`

howardchung:
@nicholashh ran this, we are good to merge when conflict is resolved

PR #1086, opened 2016-06-14
node 6

howardchung:
- [ ] figure out why requested parses stall

Weird issue that seems to happen with Node 6.  For some reason, parses requested via webpage appear to stall at completion (replay is downloaded, presumably parsed, but apparently no 'end' event is fired), so the job never completes.

The Travis test passes, but I'm hesitant to push this to production without figuring out why.


howardchung:
No longer needed

PR #1100, opened 2016-06-22
extract ward entityLeft events

howardchung:
Early PR for additional ward data.  Extract a list of ward entered/left events and store in a match-level property called "log".

Intentionally generic name in case we want to store different types of events later.

UI things we can add:
* Team ward coverage % (time)
* replayable ward state (scrubbable map)


howardchung:
redo PR into match_logs branch

PR #1104, opened 2016-06-25
New player api

howardchung:
Improve /api/players endpoints to return more focused results.

Cassandra is now required.


nt47:
么么哒~

PR #1132, opened 2016-07-22
implement long-running parse server

### LINKED ISSUE #1129: explore long-running java parser process
Have a long-running java process (webserver) that reads replay as POST body and returns logs as response.

Might benefit from reduced JVM startups and hot path optimization.

http://stackoverflow.com/questions/3732109/simple-http-server-in-java-using-only-java-se-api

howardchung:
fixes #1129 

Reviewers:
@spheenik


svc/parser.js
```diff
@@ -59,6 +59,16 @@ app.get('/redis/:key', function(req, res, cb)
 });
 app.listen(config.PARSER_PORT);
 //END EXPRESS
+// Start Java parse server
+spawn("java", [
```
howardchung: if this child ever exits, we probably want to restart the node process as well.

howardchung:
Make parse server port configurable via command line argument

PR #1136, opened 2016-07-25
use java thread pool

howardchung:
Thread pool to allow multiple parallel requests on parseserver

### DIFF
--- a/java_parser/src/main/java/yasp/Main.java
+++ b/java_parser/src/main/java/yasp/Main.java
@@ -13,7 +13,7 @@ public class Main {
     public static void main(String[] args) throws Exception {
         HttpServer server = HttpServer.create(new InetSocketAddress(Integer.valueOf(args[0])), 0);
         server.createContext("/", new MyHandler());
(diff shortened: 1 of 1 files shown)


gortiz:
Why not use a cached thread pool?

howardchung:
what is the advantage?

gortiz:
A cached thread executor pool has as many threads as necessary. Once a
thread finished its task, it waits a fixed number of seconds for a new one.
If there is no new task, the thread is destroyed. The main problem this
executor has is that it can potentially generate too many threads.

You could also try to use a ForkJoinPool but, to be honest, the more I
think about your use case, the more I would opt for FixedThreadPool, as it
is the simplest way to do it.

2016-07-26 11:26 GMT+02:00 Howard Chung <notifications@github.com>:

howardchung:
Yeah, that's why I opted for the fixed pool (to avoid the potential problem of creating too many threads).  In practice, we control exactly how hard we hit the parse server (current parallelism of parse jobs is 2), so I think both would work fine.

PR #1157, opened 2016-08-07
ranking update

howardchung:
* bump mmr weighting factor
* score negative points for losses

merge before october 1 ranking reset


howardchung:
Reviewers:
@albertcui 
@nicholashh

PR #1199, opened 2016-09-07
Setup

micaelbergeron:
These are the main things I found when starting to work on the code.

- Cassandra was eating 4Gb of memory, which is insane in a dev env.
- The `docker-compose up` was missing the yasp/parser image which made impossible to parse a replay after a fresh boot.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,9 @@ services:
   cache:
     image: redis:3
     network_mode: host
+  parser:
+    image: yasp/parser
+    network_mode: host
   web:
     build: .
     # 'basic' to launch basic configuration as in profiles/basic.json
@@ -31,3 +34,4 @@ services:
     depends_on:
       - db
       - cache
+      - parser
--- a/docker/cassandra.dockerfile
+++ b/docker/cassandra.dockerfile
@@ -1,5 +1,9 @@
 FROM daerdemandt/cassandra-init-manual
 
+# Lower cassandra memory limits
+ENV MAX_HEAP_SIZE=128M
+ENV HEAP_NEWSIZE=24M
+
 COPY sql /tmp/sql
 RUN cp /tmp/sql/init.cql            /docker-entrypoint-init.d/10-init.cql; \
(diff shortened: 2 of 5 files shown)


docker/cassandra.dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM daerdemandt/cassandra-init-manual
 
+# Lower cassandra memory limits
+ENV MAX_HEAP_SIZE=128M
```
howardchung: Can you apply these to the non docker compose as well?  See scripts/dev.sh

docker-compose.yml
```diff
@@ -16,6 +16,9 @@ services:
   cache:
     image: redis:3
     network_mode: host
+  parser:
+    image: yasp/parser
```
howardchung: You can remove the out of date tag from the readme

docker/cassandra.dockerfile
micaelbergeron: On it.

micaelbergeron: Is the scripts/prebuild.sh used to build a production image?

howardchung: It is used for tests on Travis.  It doesn't seem to have RAM issues so probably best to leave it.

howardchung:
@Daerdemandt for review

scripts/prebuild.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
 sudo service postgresql stop
-sudo docker run -d --privileged --name cassandra --net=host -e "CASSANDRA_LISTEN_ADDRESS=127.0.0.1" cassandra:3
+sudo docker run -d --privileged --name cassandra --net=host \
```
howardchung: I would revert this to make Travis match production as much as possible.

Daerdemandt:
Looks fine, works fine. Can't say anything about Travis though.

howardchung:
CI looks good, but I'd still revert the change to previous script.

PR #1200, opened 2016-09-08
add items to db

howardchung:
- [x] run schema edit


nicholashh:
`create table items (id int PRIMARY KEY, name text, cost int, secret_shop smallint, side_shop smallint, recipe smallint, localized_name text);`

PR #1215, opened 2016-09-16
Update docker configuration

dartharaknus:


Dockerfile
```diff
@@ -2,14 +2,12 @@ FROM node:6.5.0
 
 ENV NPM_CONFIG_LOGLEVEL warn
 
-WORKDIR /usr/src/yasp
+COPY . /usr/src/yasp
 
-ADD package.json /usr/src/yasp
-RUN npm install
+WORKDIR /usr/src/yasp
```
howardchung: The original intention of splitting this was implemented by @paralin and is meant to allow caching of the node_modules.  I've noticed that it doesn't seem to be working on Travis lately, so perhaps it can be removed.

Dockerfile
```diff
@@ -2,14 +2,12 @@ FROM node:6.5.0
...
 
-ADD . /usr/src/yasp
-RUN npm run webpack
+RUN npm run build
 
-ENV PATH /usr/src/yasp/node_modules/pm2/bin:$PATH
+ENV PATH /usr/src/yasp/node_modules/.bin:$PATH
```
howardchung: why is this path changed?

README.md
```diff
@@ -16,35 +16,39 @@ Tech Stack
 
 Quickstart (Docker)
 ----
-* Install Docker: `curl -sSL https://get.docker.com/ | sh`
-* Start containers and initialize databases: `sudo bash scripts/dev.sh`
+
+### Running the server
+
+* Install Docker: `curl -sSL https://get.docker.com/ | sh`. If you are on windows, make sure you shared the working drive with docker.
```
howardchung: capitalize Windows and Docker

README.md
```diff
@@ -16,35 +16,39 @@ Tech Stack
...
 * Create .env file with required config values in KEY=VALUE format (see config.js for a full listing of options) `cp .env_example .env`
   * `STEAM_API_KEY` You need this in order to access the Steam Web API.  
   * `STEAM_USER, STEAM_PASS` The retriever requires a Steam account in order to fetch replay salts. It is recommended to use a new account for this purpose (you won't be able to log into the account while the retriever is using it).  If you don't care about getting replay salts/downloading replays then you can skip this step.
-* Get a terminal into the running container: `sudo docker exec -it yasp bash`
-* Rebuild inside the container (your local directory hides the built files): `npm run build`
+* Start containers and initialize databases: `sudo docker-compose up`. If you have updated package.json, please use `sudo docker-compose up --build`
```
howardchung: Shouldn't the user modify package.json once in the container, then rebuild?  I don't see a use case where they edit package.json outside the container and then have to build it.

config.js
```diff
@@ -40,9 +40,12 @@ var defaults = {
     "SCANNER_PORT": "5500",
     "PARSER_HOST": "http://localhost:5600", //host of the parse server
     "POSTGRES_URL": "postgresql://postgres:postgres@localhost/yasp", //connection string for PostgreSQL
+    "POSTGRES_TEST_URL": "postgresql://postgres:postgres@localhost/yasp_test",
```
howardchung: since these variables are only used for test, I kind of liked having them defined in the test block only.

docker-compose.yml
```diff
@@ -3,35 +3,57 @@
...
     ports:
-      - "5000"
-    volumes:
-      - .:/usr/src/yasp
-    network_mode: host
-    depends_on:
-      - db
-      - cache
+      - "5000:5000"
+    environment:
```
howardchung: Can we just use network mode host and avoid having to redefine all this config?

docker-compose.yml
```diff
@@ -3,35 +3,57 @@
...
+      REDIS_TEST_URL: redis://odota-redis:6379/1
+      CASSANDRA_URL: cassandra://odota-cassandra/yasp
+      CASSANDRA_TEST_URL: cassandra://odota-cassandra/yasp_test
+    links:
+      - postgres
+      - cassandra
+      - redis
       - parser
+    volumes:
+      - ./public/:/usr/src/yasp/public
```
howardchung: why do we need to individually list directories here?

howardchung: i see, this is the mapping of local directories into the Docker container.  My concern is that in the future some developer will try to add a directory and get confused as to why it's not working.

docker-compose.yml
```diff
@@ -3,35 +3,57 @@
...
+      REDIS_URL: redis://odota-redis:6379/0
+      REDIS_TEST_URL: redis://odota-redis:6379/1
+      CASSANDRA_URL: cassandra://odota-cassandra/yasp
+      CASSANDRA_TEST_URL: cassandra://odota-cassandra/yasp_test
+    links:
+      - postgres
+      - cassandra
+      - redis
       - parser
+    volumes:
```
howardchung: will edits to . get mirrored to the container?

docker/main-launch.sh
```diff
@@ -3,8 +3,7 @@
 # main-launch.sh
 # Script run in the main container
 # If this script exits, so does the main container
-
-npm run build
+npm run webpack
```
howardchung: why do we run webpack here?  Isn't it done as part of the build?

Dockerfile
alexbarrett: This is probably the right thing. All binaries provided by installed npm packages are available in `node_modules/.bin` including `pm2`. This means you'll also be able to run tools like `mocha`, `uglifly`, `webpack` etc. directly from the path as well as `pm2`.

README.md
alexbarrett: We should remove all the references to `sudo docker`; shouldn't be running things as root. Instead these commands should be run by someone in the `docker` user group (which you can add your own account to).

docker-compose.yml
alexbarrett: I didn't use the changes in this PR, but removing network_mode: host was the main thing I had to do to get OpenDota in Docker working on Windows. Might be possible to get it working with it, but I'm curious what the motivation was for using host was in the first place?

howardchung: Mainly it's to avoid the need to manage port forwarding.  I've also heard that there's a performance penalty for using Docker network stack instead of host's.

dartharaknus: I had to do this to get this working on both windows and osx. network mode host is fucking with port forwarding.

config.js
dartharaknus: Since I have to do [this](https://github.com/odota/core/pull/1215#r79283161), we can't access the other services using `localhost` which means we need to provide their URLs in `docker-compose.yaml`. For the tests to run in docker container properly, I had to do this.

README.md
dartharaknus: @howardchung When they pull changes from github which has edited `package.json`.

docker/main-launch.sh
dartharaknus: Since we are mounting `public` dir, the `public/build` directory will be overwritten. Which means we have to run webpack here.

docker-compose.yml
dartharaknus: No, only edits to the given folders will be mirrored to container. I have included all the folders which are considered non-startup and mounted them. Any other changes means you have to rebuild the container.

docker-compose.yml
dartharaknus: I had to do this because if we mount the whole repo directly, then the `node_modules` will get overwritten and we have install them everytime when we run a container instead of when building the container only.

Regarding the issue with new folders, I am going to assume that it's going to be very rare.

howardchung: Can we just mount the whole repo?  The user will need to rebuild once, but any further operations can be incremental.

dartharaknus: We can do it like that, but it's not the docker way of doing things. There should be no other 1 time commands except the `docker-compose up`

PR #1244, opened 2016-09-24
Use alpine-node in docker

dartharaknus:


dartharaknus:
Nvm this. It has issues on windows.

PR #1263, opened 2016-10-02
Fix kda

blukai:


blukai:
That's a good question. Actually it would be nice if we could refresh kda in old matches. But that's not a big deal I guess.

howardchung:
Once the code is deployed, requesting the match (reinserting) will update the field.  However, it would be pretty expensive to run a task to rewrite the KDA values for about 5 billion rows.

mike-robertson:
If we are turning deaths into 1 if it is 0, it is also skewing data since it isn't an accurate kda. Kda should just be infinity if you have 0 deaths, right?

howardchung:
It looks like JS calls it infinity, but really it's just undefined and we shouldn't be doing it.  Therefore we have the options of either statically adding 1 (making it a Kills to Lives ratio, this is the current behavior) or accepting this PR and converting 0 to 1.

blukai:
KDA cannot be infinity. K + A / D, if D = 0 => K + A

dartharaknus:
The thing is we know, D=0 is better than D=1, but adding 1 to D is not the good way to do or show this. I think it's better if we do `max(D, 1)` and put a star or something beside the value on the frontend if D=0

howardchung:
After talking with the Standard Deviants (a group of Dota statspeople), I think the best approach is to relabel the stat as KLA, and keep the existing formula.  We should also use it for the client-side computation as well.

dartharaknus:
@howardchung Then let's do the renames.

PR #1295, opened 2016-10-20
XP graph label in format: XP - Level L

vamo89:
Code with a little duplication, since this will be thrown away in a few weeks.
I tried to use the function from utility but I should be missing something.
require('utility') from generateCharts doesn't work, maybe because it's in another folder. I tried some variations as well but without success.

I'd be glad if someone know how to fix that. Otherwise, maybe it's better to accept just the generateCharts change.

### DIFF
--- a/public/js/generateCharts.js
+++ b/public/js/generateCharts.js
@@ -1,3 +1,5 @@
+const constants = require('dotaconstants');
+
 window.generateCharts = function generateCharts(data) {
     var color_array = [];
     for (var key in constants.player_colors) {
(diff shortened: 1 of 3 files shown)


public/js/generateCharts.js
```diff
@@ -77,9 +83,18 @@ window.generateCharts = function generateCharts(data) {
...
+                    return this.getTooltipContent(d, defaultTitleFormat, chart.valueFormat || defaultValueFormat, color);
                 }
             }
         });
     });
-};
\ No newline at end of file
+};
+
+function getLevelFromXp(xp) {
```
howardchung: If you want to use utility you need to give the relative path: '../../../util/utility'.  Optional since this code is about to be deprecated.

Issue #1047, opened 2016-05-21
mmstats page is reporting wrong numbers


rossengeorgiev:
Hello, I've spotted an error on the MM Stats page.  
If you look at the page now it is showing 0 for Singapore(SEA) and Brazil (South America). That wrong, it should be China and South Korea. The region with most players searching should be SEA, followed by EU West.

Currently the mmstats message would return an array like this:

`[490,1966,8725,7607,0,1901,0,8955,7085,181,221,1014,548,234,68,179,110,807,1009,581,1139]`

Notice the 0's are at index `4` and `6`. See `matchgroups.txt` below for the mappings:

https://github.com/SteamDatabase/GameTracking/blob/master/dota/game/dota/pak01_dir/scripts/matchgroups.txt

Also, the page should use the names from `matchgroups.txt`, since those are the ones displayed in the client, rather than the region names.

howardchung:
Thanks.  Unfortunately I have very little context on this.  @albertcui is out until August.  We can wait for him to come back, or we can accept a PR that fixes this issue.

howardchung:
Moving to backlog since @albertcui probably won't have time to fix this in the near future.  We may retire the mmstats feature.

delfwag:
maybe just retire per country stats? the general mmr stats are still somewhat useful i feel.

howardchung:
This is a separate issue, it's the mmstats page that shows currently queuing players in each region.

delfwag:
it still gives an idea of the general trends of region specific population depending on time, maybe just have a message indicating its only for general trend information and not fully working for now?

howardchung:
I haven't ported this feature to the new UI yet.  If @albertcui chooses to do so he should also fix this issue, otherwise I think we will retire it.

rossengeorgiev:
@delfwag not really

Compare https://www.opendota.com/mmstats and https://dota.rgp.io

howardchung:
We are retiring this feature.  Use https://dota.rgp.io for all your MMStats needs :)

PR #1318, opened 2016-10-29
add mmstats api

albertcui:


howardchung:
I don't think we should merge this unless we fix https://github.com/odota/core/issues/1047

howardchung:
Suggest opening a new PR if we want this.

Issue #815, opened 2016-01-06
Track alchemist aghs


waitwhereami:
Thought it would be a nifty feature to be able to see who received an Aghanim's Scepter from Alchemist. We can currently see how many he purchased over the course of a match, but since they do not show in the inventory of the receivers, it seems to be impossible to determine who got one from the current data.

TODO:
We may want to dump the existing data (we don't do anything with it) and start only tracking modifiers applied to heroes.

Can also create a new field modifier_received.

Candidates:
modifier_item_ward_true_sight (doesn't seem to be present?)
modifier_item_dustofappearance (looks like this is applied to all units in range, not just invis ones.  That limits its usefulness)
modifier_item_ultimate_scepter_consumed (applied when receiving aghs from alchemist?)
modifier_legion_commander_duel_damage_boost (received when winning a duel, seems likely to only be applied once though)
modifier_silencer_int_steal
modifier_pudge_flesh_heap

List:
https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/Scripting/Built-In_Modifier_Names

howardchung:
It might be possible to look at modifier_applied to determine who received the buff.

howardchung:
So currently we have modifier_applied for each player, which tracks the modifiers applied by a player and the count of that modifier applied, but not WHO it was applied on.  So to support this we would need to add modifier_received to each player which counts the modifiers received by each player.

howardchung:
I think what we can start with is making a whitelist of modifiers applied we care about, and display that data.  We could report how many times alchemist granted the aghs buff (although not to whom).

Stuff like LC duel might also be able to be counted.

howardchung:
Status: Code in processExpand adds this modifier to the modifiers_applied object.  We'll probably wait until after kube migration, roll out the update to parse nodes, then build the frontend to display it and roll that out.

TrentBartlem:
Latest [patch](http://store.steampowered.com/news/24295/):

>  Post Game screen changes:  
- Added a column for Permanent Buffs. This currently tracks Pudge’s Flesh Heap, Legion Commander’s Duel Damage, Silencer’s Intelligence Stolen, Consumed Moon Shard and Consumed Aghanim’s Scepter.

Hopefully there's API support for that.

howardchung:
I took a quick glance at the GetMatchDetails call and it does not seem to return this data.  Perhaps @Crazy-Duck can comment on progress toward implementing this call in node-dota2.

[...]

PR #1322, opened 2016-10-30
store party data in cassandra

howardchung:
API change: UI will be need to be updated to handle the change to party data format.

Schema updates:
- [ ] psql `ALTER TABLE match_gcdata DROP parties;`
- [x] psql `ALTER TABLE player_matches ADD party_id int;`
- [x] psql `ALTER TABLE player_matches ADD permanent_buffs json[];`
- [x] cqlsh `ALTER TABLE player_matches ADD party_id int;`
- [x] cqlsh `ALTER TABLE player_matches ADD permanent_buffs text;`


howardchung:
Current approach only stores party data if we parse the match.  Need to rethink how we insert it into Cassandra (maybe an insertMatch call directly from getGcData).  Solution should also handle permanent_buffs: #815 )

PR #1388, opened 2016-12-26
Hero stats endpoint

howardchung:
This replaces the existing publicHeroes endpoint which was not in production use and therefore comes with a semver bump to v11.0.0

Contains data for pro pick/ban/win rates and public pick/win rates in one call.

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1397,7 +1397,11 @@ Please keep request rate to approximately 1/s.
                 objectResponse[row.hero_id] = Object.assign({}, objectResponse[row.hero_id], row);
               });
             });
-            return res.json(Object.keys(objectResponse).map(key => objectResponse[key]));
+            return res.json(Object.keys(objectResponse).map(key => {
+              // add hero_id prop to each object that is a copy of id
+              const heroId = objectResponse[key].id;
+              return Object.assign({}, objectResponse[key], { hero_id: heroId });
+            }));
           });
         },
       },
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1302,11 +1302,11 @@ Please keep request rate to approximately 1/s.
         },
       },
     },
-    '/publicHeroes': {
+    '/heroStats': {
       get: {
         summary: 'GET /',
-        description: 'Get list of heroes in public matches',
-        tags: ['public heroes'],
+        description: 'Get stats about hero performance in recent matches',
+        tags: ['hero stats'],
         parameters: [
           params.minMmrParam,
           params.maxMmrParam,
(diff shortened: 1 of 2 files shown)


routes/spec.js
```diff
@@ -1324,39 +1324,87 @@ Please keep request rate to approximately 1/s.
...
+            ORDER BY hero_id
+          `, [minTime, maxTime])
+            .asCallback(cb);
+            },
+          }, (err, result) => {
+            if (err) {
+              return cb(err);
+            }
+            // Build object keyed by hero_id for each result array
+            const objectResponse = JSON.parse(JSON.stringify(constants.heroes));
```
mike-robertson: Is there a reason to do a deep copy? Also not sure serializing and deserializing is the most efficient way to handle it.

routes/spec.js
```diff
@@ -1324,39 +1324,87 @@ Please keep request rate to approximately 1/s.
...
+            Object.keys(result).forEach((key) => {
+              result[key].rows.forEach((row) => {
+                objectResponse[row.hero_id] = Object.assign({}, objectResponse[row.hero_id], row);
+              });
             });
+            return res.json(Object.keys(objectResponse).map(key => objectResponse[key]));
```
mike-robertson: This also seems unnecessary

routes/spec.js
```diff
@@ -1324,39 +1324,87 @@ Please keep request rate to approximately 1/s.
...
-          AND avg_mmr > ?
-          AND avg_mmr < ?
-          AND start_time > ?
-          AND start_time < ?
-          ORDER BY match_id desc LIMIT 10000) 
-          limited_public_matches USING(match_id)
-          WHERE hero_id > 0
-          GROUP BY hero_id
-          ORDER BY hero_id
+          async.parallel({
```
mike-robertson: Idk much about this async library, but I usually prefer to deal with promises instead of callbacks. That would require change a lot of stuff, but it might be worth doing at some point.

howardchung: How do promises work with arrays?  e.g. if I have an array of ids and I want to do a db lookup for each and return an array of resolved db objects

routes/spec.js
howardchung: I don't want to mutate the original constants.heroes object.

mike-robertson: Instead of mutating you could just build a new object. So keep the left side of 1397 the same, but  make the right side use the original object. Then nothing is mutated before the response is sent.

howardchung: do I need to pass constants.heroes[row.hero_id] to Object.assign()?  I feel like I shouldn't need to do it multiple times.

mike-robertson: I'm not 100% sure I understood that. You already do pass that multiple times to object assign, it's just that you cloned the object so you use the cloned object.

howardchung: Actually, I think I need a copy since I need to ensure every hero exists in the output array.

routes/spec.js
```diff
@@ -421,7 +421,7 @@ Please keep request rate to approximately 1/s.
     },
     '/players/{account_id}': {
       get: {
-        summary: 'GET /',
+        summary: 'GET /players',
```
dartharaknus: Why doesn't this have `account_id`?

routes/spec.js
```diff
@@ -237,7 +237,7 @@ Please keep request rate to approximately 1/s.
   paths: {
     '/matches/{match_id}': {
       get: {
-        summary: 'GET /',
+        summary: 'GET /matches',
```
dartharaknus: Why doesn't this have `match_id`?

routes/spec.js
dartharaknus: Async library is pretty cool and simple enough. No need to complicate this with promises.

routes/spec.js
```diff
@@ -1324,39 +1324,87 @@ Please keep request rate to approximately 1/s.
...
-          AND start_time < ?
-          ORDER BY match_id desc LIMIT 10000) 
-          limited_public_matches USING(match_id)
-          WHERE hero_id > 0
-          GROUP BY hero_id
-          ORDER BY hero_id
+          async.parallel({
+            publicHeroes(cb) {
+              db.raw(`
+            SELECT
```
dartharaknus: Bad indentation for this statement.

dartharaknus: Similarly below.

routes/spec.js
dartharaknus: I agree with mike. Use another object instead of deep copying.

mike-robertson:
I don't think we should switch to promises in the context of this, but I disagree with your statement @dartharaknas. Promises require a higher initial understanding (of promises), but it's much more understandable and maintainable than callback hell.

dartharaknus:
That's the thing. I don't understand this callback hell. I feel like callbacks are better understandable and maintainable, but I guess people can have different opinions.

mike-robertson:
This is a simple example, but I think it illustrates my biggest issue with it. Keep in mind this async library should avoid some of these kinds of issues, though.
```js
asyncCall((err, res) => {
  if (err) handle(err);
  secondCall((err, res) => {
    if (err) handle(err);
    thirdCall((err, res) => {
      if (err) handle(err);
      done(res);
    }
  })
})
```
vs.
```js
asyncCall()
  .then(res => secondCall())
  .then(res => thirdCall())
  .then(res => done(res))
  .catch(err => handle(err))
```
This async library is trying to solve what promises do (with some rxjs like functional behavior) before promises became the preferred solution. Plus promises are native now, and if people like async/await (I personally don't for most stuff), that also uses promises.

Regardless though, swapping to use that would be a large undertaking (probably need to happen piecemeal) and wouldn't be a decision to make lightly or during this PR of course.

routes/spec.js
mike-robertson: @howardchung You could avoid it like this:
```js
const response = Object.keys(result).reduce(
  (obj, key) => Object.assign(obj, constants.heroes[key]),
  {}
);
```
or better (not needing to do the object to array conversion):
```js
const response = Object.keys(result).reduce(
  (arr, key) => arr.concat(constants.heroes[key]),
  []
);
```

howardchung:
How do we get all the results of the async calls in scope for the final assembly of the JSON data though using promises?  The async library has utility functions that can take an array/object of async functions, and calls back to a final function which gets the array/object with the results of the async calls.

I agree that nesting too many levels of callbacks is bad, but that's why I added the async library--if I see a function get more than 2 levels deep or so I rewrite it to use the library to flatten the structure.

routes/spec.js
howardchung: The keys of result aren't the hero_ids, though, they're the properties of the input object (publicHeroes, proHeroes, proBans).  Each is an array of data, where each element is one row of the returned SQL response.  We have no guarantee that these arrays contain every hero, since if a hero does not appear at all (has a count of 0), GROUP BY will not include it.

mike-robertson:
Promises can do that as well using [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all). Here's an example:
```js
Promise.all([call1(), call2(), call3()])
  .then(values => values.map(doSomething))
  .catch(err => handle(err));
```
Anyways, like I said it seems this async library is a 'how to handle callback hell' before promises became the standard. But I don't see a reason to rewrite it unless you wanted a more 'standard' control flow. Promises are also faster, but we would have to use a library called bluebird which is surprisingly faster than the native implementation. [Here are performance benchmarks](http://bluebirdjs.com/docs/benchmarks.html).

routes/spec.js
mike-robertson: Ah, that makes sense, I misread that. Still, a deep copy seems like overkill here: we would be better off doing a shallow copy since we aren't mutating deep properties, right?

```js
let heroes = { ...constants.heroes };
const response = Object.keys(result)
  .forEach(key => {
    heroes = result[key].rows.reduce(
      (obj, row) => Object.assign({}, obj, {
        [row.hero_id]: row,
      }),
      heroes,
    );
  });
```

mike-robertson: I see, that makes sense. My bad for misreading that

dartharaknus: We can still avoid deep copying a bit. But I guess it's okay. I am just worried about the performance issues due to deep copying.

And can't we cache this endpoint?

howardchung: I'll add caching if there are performance issues.  However we'd probably only be able to cache the "default/no query string" resposne since we allow filtering by time/MMR.

PR #1403, opened 2017-01-26
update heroStats API

howardchung:
* halve benchmark retention time (save that Redis memory!)
* add bucketing to heroStats API by MMR
* heroStats now returns data for current month (may add support for fetching historic months if there's demand for it)
* Move heroStats computation to background task and store in Redis
* disable mmStats background task


svc/worker.js
```diff
@@ -9,7 +9,7 @@ const db = require('../store/db');
 const queries = require('../store/queries');
 const buildSets = require('../store/buildSets');
 const utility = require('../util/utility');
-const getMMStats = require('../util/getMMStats');
+// const getMMStats = require('../util/getMMStats');
```
dartharaknus: Why not delete the code instead of commenting? It will always be available in git history if you want to retrieve it later anyway.

howardchung: For things we "might" add back at some point I like to have a visual reminder.  I don't know if @albertcui will ever bring this back, though.

PR #1414, opened 2017-02-25
Roaming lane [WIP]

### LINKED ISSUE #1411: Add roaming to lane position
Should be easy to figure out. If the ratio of the mode to the rest is at some threshold (close to 1:1) the hero is considered roaming.

Thoughts?

masad-frost:
resolves #1411

### DIFF
--- a/sql/create_tables.cql
+++ b/sql/create_tables.cql
@@ -139,6 +139,7 @@ CREATE TABLE player_caches (
   neutral_kills text,
   lane text,
   lane_role text,
+  is_roaming text,
   obs text,
   sen text,
   item_uses text,
--- a/util/utility.js
+++ b/util/utility.js
@@ -292,6 +292,10 @@ function modeWithCount(array) {
   return { mode: maxEl, count: maxCount };
 }
 
+function mode(array) {
+  return modeWithCount(array).mode;
+}
+
 /**
  * Determines if a match is significant for aggregation purposes
  **/
@@ -707,6 +711,12 @@ function getLaneFromPosData(lanePos, isRadiant) {
     });
   });
   const { mode: lane, count } = modeWithCount(lanes);
+  /**
+  * Player presence on lane. Calculated by the count of the prominant
+  * lane (`count` of mode) divided by the presence on all lanes (`lanes.length`).
+  * Having low presence (<45%) probably means the player is roaming.
+  **/
+  const isRoaming = (count / lanes.length) < 0.45;
 
   // Roles, currently doesn't distinguish between carry/support in safelane
   // 1 safelane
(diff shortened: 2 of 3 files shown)


masad-frost:
I'm not sure if this is the right approach. I'm thinking maybe we should leave the lane intact and return a roaming bool i.e `getLaneFromPosData`  -> `{ lane, lane_role, is_roaming }`

howardchung:
yeah, I agree that would be less risky, since we retain all of the current data and can choose whether to use the new property or not.  It'll require a schema update though.

masad-frost:
Yeah I figured. And on the UI, we can do Mid (Roaming) or something, which gives more information

masad-frost:
Then again, it will add complexity to filter and counts if we decide to include roaming in there

masad-frost:
I added player is roaming to player caches for future filtering.

`ALTER TABLE player_caches ADD is_roaming text;`

sql/create_tables.cql
```diff
@@ -139,6 +139,7 @@ CREATE TABLE player_caches (
   neutral_kills text,
   lane text,
   lane_role text,
+  is_roaming text,
```
howardchung: would this also be useful in the SQL database for pro matches?

util/utility.js
```diff
@@ -273,9 +273,9 @@ function mergeObjects(merge, val) {
 }
 
 /**
- * Finds the mode of the input array
+ * Finds the mode and its occurances of the input array
  **/
-function mode(array) {
+function modeWithCount(array) {
```
howardchung: could we preserve the existing function API?  I like having a "mode" function that just returns the mode.  We could make it take an options object to return the additional data.

util/utility.js
```diff
@@ -726,6 +728,7 @@ function getLaneFromPosData(lanePos, isRadiant) {
   return {
     lane,
     lane_role: laneRoles[lane],
+    is_roaming: (count / lanes.length) < 0.45,
```
howardchung: maybe add a comment to explain what this is doing

sql/create_tables.cql
masad-frost: I thought only raw data (`lane_pos`) is stored on that one.

howardchung: Yeah, but we could add a field.  I suppose that it should probably be part of a different effort/PR to add the lane_role/lane/is_roaming properties for easier querying/filtering.

util/utility.js
```diff
@@ -292,6 +292,10 @@ function modeWithCount(array) {
   return { mode: maxEl, count: maxCount };
 }
 
+function mode(array) {
+  return modeWithCount(array).mode;
```
masad-frost: Is this ok? I didn't do the option because I'm usually against the same function returning 2 different types.

masad-frost:
@howardchung make sure you take care of the migration before deploying

howardchung:
I already did

PR #1415, opened 2017-02-27
use alpine-node

howardchung:
@dartharaknus previously attempted to do this in #1244 

Apparently there were issues on Windows?  Not sure if that's the case anymore.

For roles that download the image a lot (retriever/parser) using alpine can save quite a bit of bandwidth/startup time.


howardchung:
waiting for node 8 LTS to be released, will probably update the image then

PR #1417, opened 2017-03-01
add minimum match rating for rankings

howardchung:
* This change adds a minimum average match rating for matches to be considered for rankings (currently 2000).
* This reduces redis memory usage


blukai:
Maybe only registered users should participate in rankings? 🍀

PR #1434, opened 2017-04-02
initial implementation of cassandra skill

howardchung:
* Move match skill data to Cassandra (store in skill property in matches and player_caches table)
* Need to add match column to matches table (done)
* Need to run migration script `skillToCassandra`
* currently about 437 million matches in the match_skill table
* after running this that table can be dropped (about 50GB currently)
* Might want to dump the match_skill table and put it in GCS or something before we drop just to have a backup (I took a snapshot of the disk so we prob don't have to do this)

### DIFF
--- a/dev/skillToCassandra.js
+++ b/dev/skillToCassandra.js
@@ -2,12 +2,10 @@ const db = require('../store/db');
 const JSONStream = require('JSONStream');
 const queries = require('../store/queries');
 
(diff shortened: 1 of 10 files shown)


howardchung:
speed testing:
On my devbox it takes about 1 minute to migrate 100,000 matches so at 437 million matches it will take about 3 days to do the migration.

PR #1446, opened 2017-04-21
moved item purchases form COMBATLOG to actions

### LINKED ISSUE #1445: handle items bought pre-game
Example:
https://www.opendota.com/matches/3122022397/purchases

Riki bought items before the game start

PavelCore:
fixes #1445 

In opendota starting items aren't shown if they were bought before the start of the game (this may happen in all pick when player picked his hero, for example). Right now, opendota gets purchases from combat_log, but combat log events appear only after beginning of the match. However, events with type of "actions" exists during picks too, so i suggest moving purchase tracking from combat log to actions. This pull request changes processExpand, so that it generate purchase events from actions events, not from combat_log.

However:
- I tested it on json generating, but not on whole project
- It doesn't get hero name by slot as i don't know how to do this, but it isn't needed, is it? (you get slot by hero name anyway) 
`  const slot = ('slot' in e) ? e.slot : meta.hero_to_slot[e.unit];`
- Right now odota/parser doesn't get which item was purchased in onSpectatorPlayerUnitOrders, so i made pull request fixing it https://github.com/odota/parser/pull/11

### DIFF
--- a/processors/processExpand.js
+++ b/processors/processExpand.js
@@ -2,7 +2,7 @@
  * Strips off "item_" from strings, and nullifies dota_unknown.
  * Does not mutate the original string.
  **/
-const item_ids = require('dotaconstants').item_ids;
+const itemIds = require('dotaconstants').item_ids;
 
 function translate(input) {
   if (input === 'dota_unknown') {
@@ -192,7 +192,7 @@ function processExpand(entries, meta) {
         type: 'xp_reasons',
       });
     },
-    DOTA_COMBATLOG_PURCHASE(e) {
+    DOTA_COMBATLOG_PURCHASE() {
       // Moved to actions due to purchases before start of the game (during picks)
 
       // purchase
@@ -214,7 +214,7 @@ function processExpand(entries, meta) {
           unit,
           key,
           type: 'purchase_log',
-        }); 
+        });
       } */
     },
     DOTA_COMBATLOG_BUYBACK(e) {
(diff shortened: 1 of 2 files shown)


processors/processExpand.js
```diff
@@ -287,6 +292,27 @@ function processExpand(entries, meta) {
       });
     },
     actions(e) {
+      // purchase
+      if (e.key === '16') {
```
howardchung: I think we should only do this for events where we don't have a PURCHASE entry since this will not work immediately for new items (we have to manually update dotaconstants).

howardchung: Perhaps we can use time or game state to determine whether it's pregame or not

howardchung:
Note I applied a quick fix to prevent the purchases in the actions object from being counted `value` times since the default behavior is to increment the object value by `value` if it exists.

https://github.com/odota/core/commit/84f2fd1b9921f60ee17194d621cdaf225971e9a5

processors/processExpand.js
```diff
@@ -176,9 +179,13 @@ function processExpand(entries, meta) {
         type: 'gold_reasons',
       });
     },
-    DOTA_COMBATLOG_GAME_STATE() {
+    DOTA_COMBATLOG_GAME_STATE(e) {
       // state
-      // we don't use this here--we already used it during preprocessing to detect game_zero
+      // check if game started
+      // not sure if this value means start of the game
```
howardchung: If you have odota/parser already you can verify this by parsing a replay and checking the output for this event

PavelCore: I checked it, it's alright. Should i rewrite it using processMetadata or just delete comment?

howardchung: Yeah, I'd move it to processMetadata so it's more consistent

PR #1449, opened 2017-04-27
Create socket service

bippum:
This PR creates a web socket server that allows users to subscribe to team IDs, league IDs, or player IDs that have signed up for OpenDota, and receive socket messages when a new match is scanned or parsed for those IDs.

### DIFF
--- a/config.js
+++ b/config.js
@@ -53,7 +53,7 @@ const defaults = {
   PUBLIC_SAMPLE_PERCENT: 10, // percent of public matches to sample in DB
   ENABLE_MATCH_CACHE: '', // set to enable caching matches in Redis
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
-  WEBSOCKET_PORT: 5000,
+  WEBSOCKET_PORT: 5000, // port for live match subscription websocket service
 };
 // ensure that process.env has all values in defaults, but prefer the process.env value
 Object.keys(defaults).forEach((key) => {
--- a/config.js
+++ b/config.js
@@ -53,7 +53,7 @@ const defaults = {
   PUBLIC_SAMPLE_PERCENT: 10, // percent of public matches to sample in DB
   ENABLE_MATCH_CACHE: '', // set to enable caching matches in Redis
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
-  WEBSOCKET_PORT: 5000, // port for live match subscription websocket service
+  WEBSOCKET_PORT: 80, // port for live match subscription websocket service
 };
 // ensure that process.env has all values in defaults, but prefer the process.env value
 Object.keys(defaults).forEach((key) => {
--- a/config.js
+++ b/config.js
@@ -53,6 +53,7 @@ const defaults = {
   PUBLIC_SAMPLE_PERCENT: 10, // percent of public matches to sample in DB
   ENABLE_MATCH_CACHE: '', // set to enable caching matches in Redis
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
+  WEBSOCKET_PORT: 5000,
 };
 // ensure that process.env has all values in defaults, but prefer the process.env value
 Object.keys(defaults).forEach((key) => {
--- a/store/queries.js
+++ b/store/queries.js
@@ -765,8 +765,7 @@ function insertMatch(match, options, cb) {
   function tellSocket(cb) {
     if (match.origin === 'scanner') {
       redis.publish('socket:matches:scanner', JSON.stringify(match));
-    }
-    if (match.origin === 'parsed') {
+    } else if (match.origin === 'parsed') {
       redis.publish('socket:matches:parser', JSON.stringify(match));
     }
     cb();
--- a/store/queries.js
+++ b/store/queries.js
@@ -762,6 +762,13 @@ function insertMatch(match, options, cb) {
     cb();
   }
 
+  function tellSocket(cb) {
+    if (match.origin === 'scanner') {
+      redis.publish('socket:matches:scanner', JSON.stringify(match));
+    }
+    cb();
+  }
+
   function decideLogParse(cb) {
     if (match.leagueid) {
       db.select('leagueid')
@@ -1138,6 +1145,7 @@ function insertMatch(match, options, cb) {
   }
   async.series({
     preprocess,
+    tellSocket,
     decideLogParse,
     upsertMatch,
     upsertMatchCassandra,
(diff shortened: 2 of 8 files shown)


svc/parser.js
```diff
@@ -201,6 +201,7 @@ function runParse(match, job, cb) {
       return cb(err);
     }
     const parsedData = createParsedDataBlob(entries, match);
+    redis.publish('socket:matches:parser', JSON.stringify(parsedData));
```
howardchung: This will fire for parse requests as well.  Is that acceptable?

svc/scanner.js
```diff
@@ -73,6 +73,7 @@ function scanApi(seqNum) {
         delayNextRequest = true;
       }
       console.log('[API] match_seq_num:%s, matches:%s', matchSeqNum, resp.length);
+      resp.forEach(match => redis.publish('socket:matches:scanner', JSON.stringify(match)));
```
howardchung: consider putting this into the insertMatch method and check for `type === 'scanner'`, that way it'll also work for backupscanner

svc/socket.js
```diff
@@ -0,0 +1,339 @@
...
+const db = require('../store/db');
+const queries = require('../store/queries');
+const buildMatch = require('../store/buildMatch');
+const async = require('async');
+
+const getPlayer = queries.getPlayer;
+const subTypes = ['player', 'team', 'league'];
+
+const wss = new WebSocket.Server({
+  port: 5000,
```
howardchung: may want to make this configurable

svc/parser.js
bippum: Yes, as the subscriber client knows where the origin is, so the client can filter for parsed and scanned matches.

svc/socket.js
```diff
@@ -0,0 +1,339 @@
...
+    if (this._subscriptions.team.indexOf(match.dire_team_id) !== -1) {
+      found.matches += 1;
+      found.team.push(match.dire_team_id);
+    }
+    if (this._subscriptions.league.indexOf(match.leagueid) !== -1) {
+      found.matches += 1;
+      found.league.push(match.leagueid);
+    }
+    if (found.matches) {
+      buildMatch(match.match_id, (err, matchData) => {
```
howardchung: do you need to buildMatch?  Can you just forward whatever was pushed from Redis to reduce DB load?

svc/socket.js
```diff
@@ -0,0 +1,339 @@
...
+        message: { err: 'No IDs provided. Nothing to subscribe to.' },
+      }, data.nonce);
+    }
+    // check if the ids are an array, otherwise create one
+    if (!Array.isArray(data.message.ids)) {
+      if (typeof data.message.ids === 'string' || typeof data.message.ids === 'number') {
+        data.message.ids = [data.message.ids];
+      }
+    }
+    // if its a player, verify the player has an opendota account
```
howardchung: I don't think this check is necessary.  It should work regardless of whether we have a DB row for that player.

svc/socket.js
```diff
@@ -0,0 +1,339 @@
...
+  GET_SUBS(ctx, data) {
+    // just return every sub we have
+    return ctx.send({
+      type: 'GET_SUBS_ACK',
+      message: ctx._subscriptions,
+    }, data.nonce);
+  },
+};
+
+wss.on('connection', (ws) => {
```
howardchung: Maybe the client could just connect to a path like `/players/{account_id}` or `leagues/{leagueid}` and then the server would just push match blobs only?  That would reduce the complexity of the server without the need to handle subscribe/unsubscribe control messages.

This is where http might be easier?

svc/scanner.js
howardchung: In fact you could even remove the line in parser.js if you edit insertMatch since parser calls that method as well.

svc/socket.js
bippum: I constructed this service to consume it with my Discord bot, and having possibly hundreds of separate connections open for player seems ugly, A web socket was easier for me to understand as well.

config.js
```diff
@@ -53,6 +53,7 @@ const defaults = {
   PUBLIC_SAMPLE_PERCENT: 10, // percent of public matches to sample in DB
   ENABLE_MATCH_CACHE: '', // set to enable caching matches in Redis
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
+  WEBSOCKET_PORT: 5000,
```
howardchung: add a comment explaining what this is for

store/queries.js
```diff
@@ -762,6 +762,13 @@ function insertMatch(match, options, cb) {
     cb();
   }
 
+  function tellSocket(cb) {
+    if (match.origin === 'scanner') {
```
howardchung: I think match.origin === 'parser' for parsed matches, check the code to make sure (otherwise you won't get parsed data blobs)

svc/parser.js
howardchung: are you planning to move this to insertMatch?

store/queries.js
bippum: Well I removed the line from scanner in order to make it work with scanner and backupscanner, parser still sends it's blobs independently

svc/socket.js
```diff
@@ -0,0 +1,290 @@
...
+  // check if this websocket is subscribed to any of the paramaters.
+  checkMatch(match, origin) {
+    const found = {
+      matches: 0,
+      player: [],
+      team: [],
+      league: [],
+    };
+    match.players.forEach((player) => {
+      if (this._subscriptions.player.indexOf(player.account_id) !== -1) {
```
howardchung: is this significantly faster than .includes?  Shouldn't both run in O(N) time?

store/queries.js
bippum: Unless you want me to move both scanner and parser blob sending into this call (actually sounds like a better idea)

svc/socket.js
bippum: http://jsben.ch/#/WWVXN

store/queries.js
howardchung: yes, that

svc/socket.js
howardchung: I got 1149 vs 1151 ms and includes actually won

store/queries.js
```diff
@@ -766,6 +766,9 @@ function insertMatch(match, options, cb) {
     if (match.origin === 'scanner') {
       redis.publish('socket:matches:scanner', JSON.stringify(match));
     }
+    if (match.origin === 'parsed') {
```
howardchung: micro-optimization, else if so this block isn't evaluated at all if the first block triggers?

PR #1451, opened 2017-04-29
fix double purchases

PavelCore:
I used in previous pull requests game_zero, thinking that was beginning of the game, but game_zero actually means time when clock in game shows 0:00. Because of it this, items should be counted two times if they were bought in time range [-1:30 -- 0]. But sometimes items were bought twice after. I guess, this bug existed due to first item purchase(*), that's why i changed e.time to game start time when the first is smaller.

To conclude, i added pre_game to meta and use it instead of game_zero and change purchase time in actions from e.time to meta.pre_game

(*) I checked purchases of bristleback in this match https://www.opendota.com/matches/3122022397/purchases
 using these scripts 
https://gist.github.com/PavelCore/3c9c1a42fc0a9b7fc4c5ac45c03127ba
https://gist.github.com/PavelCore/c28231d236bd44ce3a948df3dc8b1b7e
and got 2 entries for each purchase (type: purchase and type: purchase_log) with both old (even before fix-start-items PR) and new processorExpand.js, but i fixed purchase time and double purchase with couriers in this match.


processors/processMetadata.js
```diff
@@ -6,12 +6,17 @@ function processMetadata(entries) {
   const slotToPlayerslot = {};
   let gameZero = 0;
   let gameEnd = 0;
+  let preGameTime = 0;
```
howardchung: Suggest gameStart for consistency

processors/processExpand.js
```diff
@@ -293,15 +293,16 @@ function processExpand(entries, meta) {
...
         // i.e. dotaconstants doesn't have this item
         if (typeof key === 'undefined') {
           expand(Object.assign({}, e, { value: 1 }));
           return;
         }
+        // we don't want to show time of purchases which was done even before pre-game
         expand({
-          time: e.time,
+          time: meta.pre_game,
           value: 1,
```
howardchung: Is there a reason we can't use time?  It'll just be a smaller negative value?

PavelCore: I don't know why, but puchases were doubled even not in pre-game: that last till approx 5 min and then problem disappears even both scripts (old and new) got same outputs corresponding to this items. Maybe some problem in UI -- dont know ( I just looked to dotabuff and saw that they use time max(time, -1:30)).

PavelCore: sounds strange, but i really don't know, trying to find the problem right now

howardchung:
might be a lint issue, otherwise LGTM

PR #1463, opened 2017-05-16
greevils greed 7.06

evaldobratti:
it seems that my previous impl was wrong in the base gold :P


dartharaknus:
What happens when you are viewing previous patch matches with this code? Shouldn't we do patch based calculations?

howardchung:
I think the gold amounts are locked once the match is parsed.

PR #1470, opened 2017-05-28
RFC: retriever: implement two-factor and SteamGuard authentication

aiedail92:


howardchung:
node-dota2 actually has support for Steam Guard auth already, although I'm not sure if it handles 2FA.  You might want to make your pull request there instead so more users can benefit from the change.  That implementation stores the data in a local filesystem file instead of Redis. @Crazy-Duck

Crazy-Duck:
yes, node-dota2 does indeed already support 2FA, because node-steam supports it :)

svc/retriever.js
```diff
@@ -8,6 +8,8 @@ const Dota2 = require('dota2');
 const async = require('async');
 const express = require('express');
 const cp = require('child_process');
+const redis = require('../store/redis');
```
howardchung: In production, the retriever nodes run separately from the rest of the system and do not have Redis access.

aiedail92:
Yeah, 2FA and SteamGuard auth are already in node-steam, but sentry file management and of course the actual passing of the authentication token need to be handled separately.

Since redis is not accessible in the retriever in production, what alternatives are there? Would it be acceptable to have a separate redis instance?

Also, is the whole `/auth` thing acceptable? I guess it would have problems if the retriever service gets scaled, but I don't know exactly what production looks like.

howardchung:
You can store the data in a filesystem file, maybe, like the node-dota2 examples.  That won't work for multiple accounts though, unless node-dota2 is changed to write the sha_sentryfile to one file per account, or extend the format to support multiple accounts.

The node-dota2 examples rely on passing it through a config option.  Maybe that'd work for this as well?  

https://github.com/Arcana/node-dota2/blob/staging/examples/example.js#L182

aiedail92:
Saving the sentry to the file system wouldn't work if you scale the service (unless maybe playing with docker volumes?)

Passing auth codes in the config file IMHO is not a great option: you'd have to start the service, collect all the auth codes in the config file and restart the service. Plus, you'd have to do that every time a client disconnects and needs to re-auth for whatever reason. It looks nicer to me to allow authenticating the clients while the service is up, what do you think about that?

howardchung:
It works if you only run one instance per VM. That's the case in production.  However, filesystem files do not persist between instance restarts since they can get run on a pool of VMs.

Auth endpoint sounds OK to me.

This isn't relevant for production since our accounts don't use steam guard/2FA.  If you put the redis require under a feature switch/config option so it doesn't run in production, then this is OK if you need to use Steam Guard/2FA for your own instance.

aiedail92:
Okay, I'll work on it as soon as I have some spare time.

Thanks for your feedback!

aiedail92:
I have added an `ENABLE_RETRIEVER_ADVANCED_AUTH` entry in the config file (disabled by default) and made all the new code conditional on that.

Let me know if I need to change anything else!

howardchung:
LGTM

I still think there's a decent case to be made for moving some of this functionality to the node-dota2 library.  The choice of storage backend (redis or file) should be left up to the user, but the internals of computing the SHA hash etc should be library code.

PR #1497, opened 2017-07-14
Use DOTA_COMBATLOG_TEAM_BUILDING_KILL instead of CHAT_EVENT_*

### LINKED ISSUE #1486: switch tower destroyed from CHAT_EVENT_TOWER_KILL to DOTA_COMBATLOG_TEAM_BUILDING_KILL
We can maintain an array/dictionary tracking the lifestate of each tower entity during the parse, and therefore use that to emit tower destroyed events that are aware of which tower was destroyed, rather than relying on the chat event which only tells us the time but not which tower.

rynobax:
fixes #1486


howardchung:
LGTM

@mdiller we probably want to update the story code to support the new data before merging this since it removes the old events.

howardchung:
Do we also want to include shrines and the ancient here?

rynobax:
Are those "npc_dota_badguys_fillers" and "npc_dota_badguys_fort"?

howardchung:
fillers are the base buildings.  Shrines are "healers" I think (find a replay with a destroyed shrine to confirm)

howardchung:
and yes fort is ancient

rynobax:
Yeah I can add them

rynobax:
The test failed but it seems like it failed initializing a table unrelated to the changed code.,,

howardchung:
Yeah, I'm fixing that issue now.  Sorry about that

processors/processExpand.js
```diff
@@ -121,6 +121,22 @@ function processExpand(entries, meta) {
       const unit = e.sourcename;
       const key = computeIllusionString(e.targetname, e.targetillusion);
 
+      // If it is a building kill
+      if (e.targetname.indexOf('_tower') > -1 ||
+          e.targetname.indexOf('_rax_') > -1 ||
+          e.targetname.indexOf('_healers') > -1 ||
+          e.targetname.indexOf('_fillers') > -1 ||
```
howardchung: probably remove fillers, I don't think they really count as objectives

rynobax:
I removed the fillers

howardchung:
Holding until the GC/salt fetching issues get better

processors/processExpand.js
```diff
@@ -121,6 +121,21 @@ function processExpand(entries, meta) {
...
+          e.targetname.indexOf('_rax_') > -1 ||
+          e.targetname.indexOf('_healers') > -1 ||
+          e.targetname.indexOf('_fort') > -1) {
+        expand({
+          time: e.time,
+          type: 'building_kill',
+          unit,
+          key,
+        });
+        return;
```
howardchung: Remove this return, we still want to send this data to the killed object

PR #1499, opened 2017-07-20
Adding cluster to recent matches

rynobax:


routes/spec.js
```diff
@@ -1013,6 +1013,10 @@ Please keep request rate to approximately 3/s.
                     description: 'is_roaming',
                     type: 'boolean',
                   },
+                  cluster: {
+                    description: 'cluster',
+                    type: 'number',
```
howardchung: It's been pointed out that this should be 'integer' not number

howardchung:
Lint issue probably?

rynobax:
I think it might have been network issues.

PR #1500, opened 2017-07-21
Fix array type definition

Seraphli:
There are several places in the code where type `array` is not followed with `items` declaration. So I added definition according to this response https://api.opendota.com/api/matches/3170212000. 
NB: I change the type of `permanent_buffs` to object since its value is `null`, not `[]`.

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -560,7 +560,13 @@ Please keep request rate to approximately 3/s.
                       },
                       runes: {
                         description: 'runes',
-                        type: 'object',
+                        type: 'array',
+                        items: {
+                          type: 'object',
+                          additionalProperties: {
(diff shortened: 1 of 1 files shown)


routes/spec.js
```diff
@@ -454,7 +527,7 @@ Please keep request rate to approximately 3/s.
                       },
                       permanent_buffs: {
                         description: 'permanent_buffs',
-                        type: 'array',
+                        type: 'object',
```
howardchung: I think this should be left array

howardchung:
`npm run lint` to check the lint rules

Seraphli:
I fix the bugs. I'm not familiar with javascript, so it took me a while to figure out.

Issue #726, opened 2015-10-29
Aegis use counting as death


Renhaka:
In this match: http://yasp.co/matches/1899260771/

The shadow fiend's aegis uses are counted as deaths for the purposes of team-fights and the who-killed-who table, however is not included in the overview (as I believe it should be)

howardchung:
Yes, this is a long-standing issue with reading data from combat log.

Overview gives data from Steam WebAPI.  It matches game scoreboard.

However most of the parsed data is from combat log.  This does not differentiate between aegis/reincarnation deaths and regular deaths.

I haven't been able to think of a clean way to implement detecting aegis/reincarnation yet, but ideas are welcome.

howardchung:
Ideas welcome, suggested on Reddit:
 m_lifeState property of CDOTA_Unit_Hero

Renhaka:
I haven't been able to dig deeper and look, but I wrote a quick list of telltale signs that a death might have been utilizing as an aegis.

1) Respawn Timer is exactly 5 seconds
	a) Possible false positive exists with a hero with 19 bloodstone charges at level 15
		(Respawn Timer of (5 + 3.8 * 15) - (3 * 19) = 5)
2) Roshan would need to have been killed within 5 minutes of the death
3) Wraith King's ultimate is used and the conditions above have also been met

For WK's ult, the respawn timer would be exactly 3 seconds
A false positive here would be when Wraith King has a bloodstone with 7 charges at level 5 or 26 charges at level 20

I'm not sure if this is all qualifying conditions, but I'd figure I'd post these to get discussion rolling.

howardchung:
The problem with this level of specificity is that it's prone to breaking if Valve changes numbers.  I want something reliable and consistent.

LFT:
For wraith king, one way that would work more often than not would be to check if someone received the "Reincarnate" debuff. 
But it's not a guaranteed thing because it only affects enemy in a 900 radius :/

howardchung:
Yeah I'd prefer not to implement partial fixed since it might cause even
more confusion.

DoctorGester:
I'm not sure, but don't you have a gold lost on death tracker? And reincarnation deaths do not make you lose gold. Not sure how would that work on exactly 0 gold, but I imagine that's a very rare scenario.

micphi:
@DoctorGester, that's not actually that uncommon. If you quick buy with all of
your unreliable gold before death you don't lose any gold. Gold gained from a
[...]

PR #1509, opened 2017-07-29
Ignore aegis deaths

rynobax:
Fixes some of issue #726.  Tried to include comments with what everything does, let me know if anything is confusing.  I ran some diff tests on some games to check if it works, which I included below.  I think it might be easier to just run it on dev and compare games with prod though.

```javascript
When run on a game with no roshan kills there was no diff

Here is an example of a diff from a game where a ta died once with an aegis:
// Gyro kills
  -"hero_kills": 6,
  +"hero_kills": 5,
    -"npc_dota_hero_templar_assassin": 2,
    +"npc_dota_hero_templar_assassin": 1,
      -"key": "npc_dota_hero_templar_assassin",          "time": 1250        },        {

// TA Killed by
    -"npc_dota_hero_gyrocopter": 2,
    +"npc_dota_hero_gyrocopter": 1,
// Teamfights
  -"deaths": 9,
  +"deaths": 8,
        -"npc_dota_hero_templar_assassin": 1
      -"deaths": 2,
      +"deaths": 1,
        -"88": {              "130": 1            },

Here is an example of a diff from a game a meepo died with aegis died twice:

//meepo killed by
-"npc_dota_hero_meepo": 16,        "npc_dota_hero_ogre_magi": 3,        "npc_dota_hero_rubick": 1,
+"npc_dota_hero_ogre_magi": 2,

// rubick
  "hero_kills": 10,
  "hero_kills": 9,
    "npc_dota_hero_meepo": 1,
      "key": "npc_dota_hero_meepo",          "time": 1318        },        {

// ogre magi
  "hero_kills": 8,
  "hero_kills": 7,
    "npc_dota_hero_meepo": 3,
    "npc_dota_hero_meepo": 2,
      "key": "npc_dota_hero_meepo",          "time": 1965        },        {

// teamfights:
  -"deaths": 6,
  +"deaths": 5,
  -    "deaths": 1,          "deaths_pos": {            "124": {              "128": 1            }          },
  +    "deaths": 0,          "deaths_pos": {},
  -    "killed": {            "npc_dota_hero_meepo": 1          },
  +    "killed": {},
  -"deaths": 6,      "end": 1980,      "last_death": 1965,
  +"deaths": 5,      "end": 1969,      "last_death": 1954,
  -    "deaths": 1,          "deaths_pos": {            "156": {              "140": 1            }          }, 
  +    "deaths": 0,          "deaths_pos": {},
  -    "killed": {            "npc_dota_hero_meepo": 1          },
  +    "killed": {},

```
I have never submitted a PR that has had this many commits made between when I started and finished, so let me know if I need to do some rebase stuff to keep the history clean.

### DIFF
--- a/processors/processExpand.js
+++ b/processors/processExpand.js
@@ -163,7 +163,7 @@ function processExpand(entries, meta) {
 
       if (key === aegisHolder) {
         // The aegis holder was killed
-        if (aegisDeathTime == null) {
+        if (aegisDeathTime === null) {
           // It is the first time they have been killed this tick
           // If the hero is meepo than the clones will also get killed
           aegisDeathTime = e.time;
--- a/processors/processExpand.js
+++ b/processors/processExpand.js
@@ -41,22 +41,6 @@ function processExpand(entries, meta) {
   // Used to ignore meepo clones killing themselves
   let aegisDeathTime = null;
 
-  function getAegisHolder(heroSlot) {
-    // There are two names for each hero stored in meta.hero_to_slot,
-    // so we need to get the one with the most underscores
-    const aegisHolderPossibilities = [];
-    Object.keys(meta.hero_to_slot).forEach((hero) => {
-      const slot = meta.hero_to_slot[hero];
-      if (slot === heroSlot) {
-        aegisHolderPossibilities.push(hero);
-      }
-    });
-    const ndx = aegisHolderPossibilities
-      .map(hero => hero.match(/_/g || []).length) // Get number of _'s
-      .reduce((iMax, x, i, arr) => (x > arr[iMax] ? i : iMax), 0); // Get index of most _'s
-    return aegisHolderPossibilities[ndx];
-  }
-
   const types = {
     DOTA_COMBATLOG_DAMAGE(e) {
       // damage
@@ -161,7 +145,7 @@ function processExpand(entries, meta) {
         });
       }
 
-      if (key === aegisHolder) {
+      if (meta.hero_to_slot[key] === aegisHolder) {
         // The aegis holder was killed
         if (aegisDeathTime === null) {
           // It is the first time they have been killed this tick
@@ -414,7 +398,7 @@ function processExpand(entries, meta) {
       });
     },
     CHAT_MESSAGE_AEGIS(e) {
-      aegisHolder = getAegisHolder(e.player1);
+      aegisHolder = e.player1;
 
       expand({
         time: e.time,
@@ -423,7 +407,7 @@ function processExpand(entries, meta) {
       });
     },
     CHAT_MESSAGE_AEGIS_STOLEN(e) {
-      aegisHolder = getAegisHolder(e.player1);
+      aegisHolder = e.player1;
 
       expand({
         time: e.time,
--- a/processors/processExpand.js
+++ b/processors/processExpand.js
@@ -35,6 +35,28 @@ function processExpand(entries, meta) {
     }));
   }
 
+  // Tracks current aegis holder so we can ignore kills that pop aegis
+  let aegisHolder = null;
+
+  // Used to ignore meepo clones killing themselves
+  let aegisDeathTime = null;
+
+  function getAegisHolder(heroSlot) {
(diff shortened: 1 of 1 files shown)


howardchung:
Yeah, could you do a rebase off latest master so the commits I made and merged already don't show up?

rynobax:
Do I just squash all the commits after my first one?

howardchung:
you don't have to squash the commits if you don't want to, the rebase is just to remove the already merged commits from the history (although squashing all your changes into one commit is probably quickest)

rynobax:
Got it down to one commit.

processors/processExpand.js
```diff
@@ -35,6 +35,28 @@ function processExpand(entries, meta) {
...
   }
 
+  // Tracks current aegis holder so we can ignore kills that pop aegis
+  let aegisHolder = null;
+
+  // Used to ignore meepo clones killing themselves
+  let aegisDeathTime = null;
+
+  function getAegisHolder(heroSlot) {
+    // There are two names for each hero stored in meta.hero_to_slot,
```
howardchung: Why is the one with more underscores the correct one?

processors/processExpand.js
```diff
@@ -134,16 +161,29 @@ function processExpand(entries, meta) {
...
-      // If it is not a suicide
-      if (e.attackername !== key) {
-        expand(Object.assign({}, e, {
-          unit,
-          key,
-          type: 'killed',
-        }));
+      if (key === aegisHolder) {
+        // The aegis holder was killed
+        if (aegisDeathTime == null) {
```
howardchung: use ===

howardchung: if you also want to handle undefined, I prefer calling out both cases explicitly and using ===, it's more verbose but easier to read.

processors/processExpand.js
```diff
@@ -368,13 +414,17 @@ function processExpand(entries, meta) {
       });
     },
     CHAT_MESSAGE_AEGIS(e) {
+      aegisHolder = getAegisHolder(e.player1);
```
howardchung: I believe each interval tick now logs the items present in the inventory.  Maybe you could use that to determine if a player is holding an aegis?

rynobax: On the tick that the death message appears the aegis does not appear in the inventory message.

processors/processExpand.js
rynobax: The two names would be something like "npc_dota_hero_templar_assassin" and "npc_dota_hero_templarassassin".  It looks like the second one is only used for item purchase during the picking phase.  The number of underscores seemed like an easy way to differentiate them.  In the case that the hero name is one word, it shouldn't matter which one gets returned.

howardchung: Instead of matching by string, perhaps match by slot?  The AEGIS events already give slot, and the hero_to_slot lookup is designed to map combat log names `npc_dota_hero_antimage` to slots.  If the slots match, consider the aegis used.  Then you don't need this new function at all?

processors/processExpand.js
```diff
@@ -134,16 +161,29 @@ function processExpand(entries, meta) {
...
       }
 
-      // If it is not a suicide
-      if (e.attackername !== key) {
-        expand(Object.assign({}, e, {
-          unit,
-          key,
-          type: 'killed',
-        }));
+      if (key === aegisHolder) {
```
howardchung: e.g. use `meta.hero_to_slot[key]` here to find the slot of the killed unit

processors/processExpand.js
rynobax: That's what I initially did, but I thought it would be more efficient to just do one conversion when the aegis was picked up, then to do a slot lookup for every DOTA_COMBATLOG_DEATH.  I can change it if you want.

howardchung: Yeah, I think it would cleaner not to have to implement a new function if possible.  I doubt the lookups would add more than a couple of milliseconds to the execution time

rynobax:
Uses slot to identify aegis holder now.

PR #1538, opened 2017-10-07
resolve #1525

### LINKED ISSUE #1525: Higher resolution avatar on /peers.
Currently the avatar URL being returned is the low resolution one. Would it be possible to change this to the higher resolution URL?

builder-247:
fixes #1525 

Higher resolution avatars on /peers

### DIFF
--- a/store/queries.js
+++ b/store/queries.js
@@ -364,7 +364,7 @@ function getPeers(db, input, player, cb) {
       }
       t.personaname = row.personaname;
       t.last_login = row.last_login;
-      t.avatar = row.avatar;
+      t.avatar = row.avatar.replace('.jpg', '_full.jpg');
       return cb(err);
     });
   }, (err) => {
--- a/store/queries.js
+++ b/store/queries.js
@@ -364,7 +364,8 @@ function getPeers(db, input, player, cb) {
       }
       t.personaname = row.personaname;
       t.last_login = row.last_login;
-      t.avatar = row.avatarfull;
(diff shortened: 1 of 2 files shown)


store/queries.js
```diff
@@ -364,7 +364,7 @@ function getPeers(db, input, player, cb) {
       }
       t.personaname = row.personaname;
       t.last_login = row.last_login;
-      t.avatar = row.avatar;
+      t.avatar = row.avatar.replace('.jpg', '_full.jpg');
```
howardchung: you should just be able to do t.avatar = row.avatarfull

howardchung: or just add a new field, t.avatarfull = row.avatarfull

store/queries.js
```diff
@@ -364,7 +364,7 @@ function getPeers(db, input, player, cb) {
       }
       t.personaname = row.personaname;
       t.last_login = row.last_login;
-      t.avatar = row.avatar;
+      t.avatar = row.avatarfull;
```
howardchung: probably better to add a new field `avatarfull` in case someone is relying on this size for avatar

builder-247: What is the full process of adding a new field?

howardchung: instead of replacing this line just add a new line under it:
`t.avatarfull = row.avatarfull`.

Optionally you'd also add it to the API docs in `routes/spec.js` but that isn't 100% required

PR #1541, opened 2017-10-08
Add /heroes/{heroId}/matches route

Luchanso:
It's my first try add odota/ui#1143 /heroes/{heroId}/matches.
Careful, I don't check this code, because I can't configure backend env in docker, maybe later I will check.

Can you help me, with this code?

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -3409,14 +3409,18 @@ Please keep request rate to approximately 3/s.
         route: () => '/heroes/:hero_id/matches',
         func: (req, res, cb) => {
           const heroId = req.params.hero_id;
-          db.select('match_id', 'duration', 'start_time', 'win', 'player_matches.account_id')
-            .from('matches')
-            .join('heroes', 'heroes.id', 'player_matches.hero_id')
-            .leftJoin('notable_players', 'player_matches.account_id', 'player_matches.account_id')
-            .where({
-              'player_matches.hero_id': Number(heroId),
-            })
-            .limit(100)
+          db.raw(`SELECT
+            matches.match_id,
+            matches.start_time,
+            matches.duration,
+            ((player_matches.player_slot < 128) = matches.radiant_win) win,
+            player_matches.account_id
+            FROM matches
+            JOIN player_matches using(match_id)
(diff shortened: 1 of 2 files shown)


routes/spec.js
```diff
@@ -3368,6 +3368,68 @@ Please keep request rate to approximately 3/s.
         },
       },
     },
+    '/heroes/{hero_id}/matches': {
+      post: {
+        summary: 'POST /heroes/{hero_id}/matches',
+        description: 'Most recent games played with this hero',
+        tags: ['request'],
```
howardchung: tag should be under hero section

routes/spec.js
```diff
@@ -3368,6 +3368,68 @@ Please keep request rate to approximately 3/s.
...
+        responses: {
+          200: {
+            description: 'Success',
+            schema: {
+              type: 'array',
+              items: {
```
howardchung: I'm creating a matchObject in my teams data PR, we should reuse that

routes/spec.js
```diff
@@ -3368,6 +3368,68 @@ Please keep request rate to approximately 3/s.
...
+                    description: 'duration',
+                    type: 'integer',
+                  },
+                  start_time: {
+                    description: 'start_time',
+                    type: 'integer',
+                  },
+                  win: {
+                    description: 'win',
+                    type: 'string',
```
howardchung: this is a boolean

routes/spec.js
```diff
@@ -3368,6 +3368,68 @@ Please keep request rate to approximately 3/s.
...
+          const heroId = req.params.hero_id;
+          db.raw(`SELECT
+            matches.match_id,
+            matches.start_time,
+            matches.duration,
+            ((player_matches.player_slot < 128) = matches.radiant_win) win,
+            player_matches.account_id
+            FROM matches
+            JOIN player_matches using(match_id)
+            JOIN heroes on heroes.id = player_matches.hero_id
```
howardchung: left join just in case the hero doesn't exist/invalid value

howardchung:
I rolled your commit into:
https://github.com/odota/core/pull/1542

Luchanso:
Thank you for review, I understood all

PR #1542, opened 2017-10-09
implement hero data endpoints

howardchung:
@Luchanso

Implements:
/heroes/{heroId}/matchups: returns games played and wins against all other heroes (for pro matches)
/heroes/{heroId}/matches: return most recent games played with this hero
/heroes/{heroId}/durations: return games played and wins for each bucket of 5 minutes
/heroes/{heroId}/players: return the players who have played the most games on this hero

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -18,6 +18,9 @@ const redis = require('../store/redis');
 const packageJson = require('../package.json');
 const cacheFunctions = require('../store/cacheFunctions');
 const params = require('./params');
+const {
+  teamObject, matchObject, heroObject, playerObject,
+} = require('./objects');
 
 const redisCount = utility.redisCount;
 const subkeys = playerFields.subkeys;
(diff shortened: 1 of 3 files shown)


Luchanso:
Okey, thank you 👍

routes/spec.js
```diff
@@ -3383,21 +3124,245 @@ Please keep request rate to approximately 3/s.
...
+                },
+              },
+            },
+          },
+        },
+        route: () => '/heroes/:hero_id/durations',
+        func: (req, res, cb) => {
+          const heroId = req.params.hero_id;
+          db.raw(`SELECT
+            (matches.duration / 300 * 300) duration_bin,
```
Luchanso: It's correct? I ask because `(x / 300 * 300) === x`

routes/spec.js
```diff
@@ -3383,21 +3124,245 @@ Please keep request rate to approximately 3/s.
...
+            count(match_id) games_played,
+            sum(case when (player_matches.player_slot < 128) = matches.radiant_win then 1 else 0 end) wins
+            FROM matches
+            JOIN player_matches using(match_id)
+            WHERE player_matches.hero_id = ?
+            GROUP BY (matches.duration / 300 * 300)`, [heroId])
```
Luchanso: `(matches.duration / 300 * 300) === matches.duration` ?

routes/spec.js
```diff
@@ -3383,21 +3124,245 @@ Please keep request rate to approximately 3/s.
...
+        func: (req, res, cb) => {
+          const heroId = req.params.hero_id;
+          db.raw(`SELECT
+            account_id,
+            count(match_id) games_played,
+            sum(case when (player_matches.player_slot < 128) = matches.radiant_win then 1 else 0 end) wins
+            FROM matches
+            JOIN player_matches using(match_id)
+            WHERE player_matches.hero_id = ?
+            GROUP BY account_id`, [heroId])
```
Luchanso: Without LIMIT?

routes/spec.js
howardchung: It's basically rounding down to the nearest bucket of 5 minutes.  It works because of integer division truncating the decimals

routes/spec.js
howardchung: I think you can't use an alias in the GROUP BY clause

routes/spec.js
howardchung: Yeah, I don't think there will be a huge issue since there aren't too many pro players (~1000), but we could add this later if it becomes a performance issue.

PR #1546, opened 2017-10-12
Include courier kill events when parsing replays

Ian-MacLeod:


processors/processExpand.js
```diff
@@ -432,6 +432,14 @@ function processExpand(entries, meta) {
         team: e.player1,
       });
     },
+    CHAT_MESSAGE_COURIER_LOST(e) {
```
howardchung: You might want to use the combat log's killed event instead, that would also tell you the player who killed the courier.

Ian-MacLeod: Ah, good idea! One problem with the killed event, however, is that it doesn't seem to include information about which team's courier was killed (unless I'm missing something). In cases where the courier is not killed by a hero, I'm not sure how to figure out which team the courier belonged to. It seems a bit awkward to capture both events and piece them together, but maybe that's best?

howardchung: You could just indicate that the player killed a courier.  Is denying couriers even possible?

The sourcename field indicates the hero who controlled the unit that killed the courier, if it wasn't a hero.

Ian-MacLeod: Sorry, I wasn't clear. My concern is if a courier is killed by a neutral creep (rare, but can happen) I'm not sure how to figure out which team the courier belonged to. I was also unsure about lane creeps and towers, but I suppose I could just search for 'goodguys' or 'badguys' in the sourcename.

howardchung: Hmm, I guess it's fine to just use CHAT_MESSAGE_COURIER_LOST if we don't need to know how made the kill.  It's possible to correlate the events but I've tended to avoid doing so since it's a bit prone to errors.

Issue #1495, opened 2017-07-12
Add support for graphql


kaushik94:
Currently open data provides only a REST api and an optional SQL query endpoint. It would be nice to also have a graphql endpoint to execute more complex queries

howardchung:
On /matches?  I could see that potentially being useful for reducing the amount of data transferred with each match request.

Not sure if graphQL provides the full range of querying that SQL does (for explorer data) though.

kaushik94:
No the /explorer endpoint. We can have another endpoint like say, /graphql-explorer and then access our data layer there. I am starting out on graphql recently, do you want me to give it a try?

howardchung:
sure, feel free to prototype something and set up a demo.

You can probably just implement it in routes/spec.js as a new endpoint

kaushik94:
For now even though we might not be able to support native graphql queries out of the box we can have a graphql wrapper or something on top of our rest endpoints. So that even though we don't graphically represent our data, we can still serve requests that assume we do.

howardchung:
maybe something like this library:
https://github.com/postgraphql/postgraphql

dartharaknus:
If anyone needs help with GraphQL, don't hesitate to ping me on Discord.

howardchung:
I think @farmasek is working on this

farmasek:
Correct, I am working on it, didn't have a lot time recently but it's WIP

bdbch:
What's the status on this? GraphQL brings a lot of useful features in for developers like GraphQL schema types, fragments, the option to query what you need and much more (it's also easier to implement from my point of view since you have an explorer at your hand).

howardchung:
I think it's better for writing web apps that consume the API, but more complicated for scripts and other automatic processes to consume GraphQL. I'm open to PRs adding the functionality, but probably won't have the bandwidth to work on GraphQL features myself.

PR #1554, opened 2017-10-27
WIP Graphql poc

farmasek:
Created basic GraphQL implementation, regarding #1495  and  https://github.com/odota/ui/issues/1211

GraphQL api is available at `/api/graphql`
GraphiQL editor is running at `api/graphiql`

example of query is
image

Schema can be now extended of more features. 

The SQL offset might not be correct.

### DIFF
--- a/routes/api.js
+++ b/routes/api.js
@@ -1,3 +1,7 @@
+const { graphiqlConnect, graphqlExpress } = require('apollo-server-express');
+const bodyParser = require('body-parser');
+const schema = require('./graphql/schema');
+
 const express = require('express');
 const playerFields = require('./playerFields');
 const filterDeps = require('../util/filterDeps');
@@ -79,5 +83,12 @@ Object.keys(spec.paths).forEach((path) => {
     api[verb](route(), func);
   });
 });
+api.use('/graphql', bodyParser.json(), graphqlExpress({ schema }));
+api.use(
+  '/graphiql',
+  graphiqlConnect({
+    endpointURL: '/api/graphql',
+  }),
+);
 
 module.exports = api;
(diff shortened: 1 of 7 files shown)


store/search.js
```diff
@@ -14,6 +14,8 @@ function findPlayer(search, cb) {
 
 function search(options, cb) {
   const query = options.q;
+  const offset = options.page ? options.page * options.pageSize : 0;
+  const pageSize = options.pageSize || 500;
```
howardchung: The pageSize needs to be capped at a value (recently changed to 250) in order to prevent timeouts

dartharaknus:
Looks great man. Ping me on discord if you want to discuss anything about graphql in odota.

routes/api.js
```diff
@@ -79,5 +83,12 @@ Object.keys(spec.paths).forEach((path) => {
     api[verb](route(), func);
   });
 });
+api.use('/graphql', bodyParser.json(), graphqlExpress({ schema }));
```
howardchung: can these be added to spec.js so they show up in the documentation?

farmasek: there will be only one endpoint, specific queries are shown on GraphiQL, its something like interactive documentation

howardchung: We should still list the graphql endpoint so visitors can see it exists in the docs

sepehr500:
If this gets implemented, we can consider adding Apollo Client to the front end. I have had some great experiences with it. Would really clean up our code.

farmasek:
My intention is to use Apollo on client side, but unfortunately i dont have time much time now, so i dont see it finished any time soon.

sepehr500:
I would be interested in taking this over. I have implemented GraphQL on an enterprise system before.

howardchung:
Sure, go for it!

howardchung:
The idea is still sound, if someone wants to do add graphql support. That said, I don't know if we'd be rewriting the web UI to use graphQL, so it would be largely academic

PR #1563, opened 2017-11-19
An attempt at implementing the filter_gamesplayed feature!

wanderrful:
Partner request for https://github.com/odota/ui/pull/1324

### DIFF
--- a/routes/params.js
+++ b/routes/params.js
@@ -145,6 +145,13 @@ module.exports = {
     required: false,
     type: 'string',
   },
+  minimumPlayedParam: {
+    name: 'games_played',
+    in: 'query',
+    description: 'The minimum number of games played, for filtering hero stats',
+    required: false,
+    type: 'integer',
+  },
   minMmrParam: {
     name: 'min_mmr',
     in: 'query',
--- a/routes/params.js
+++ b/routes/params.js
@@ -145,8 +145,8 @@ module.exports = {
     required: false,
     type: 'string',
   },
-  minimumPlayedParam: {
-    name: 'games_played',
+  havingParam: {
+    name: 'having',
     in: 'query',
     description: 'The minimum number of games played, for filtering hero stats',
     required: false,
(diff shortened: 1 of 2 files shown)


routes/spec.js
```diff
@@ -1447,6 +1448,7 @@ Please keep request rate to approximately 3/s.
             });
             const result = Object.keys(heroes)
               .map(k => heroes[k])
+              .filter(hero => !req.queryObj.having || hero.games_played >= Number(req.query.having))
```
howardchung: use queryObj in the second one too

routes/params.js
```diff
@@ -145,6 +145,13 @@ module.exports = {
     required: false,
     type: 'string',
   },
+  minimumPlayedParam: {
+    name: 'games_played',
```
howardchung: if you named it 'having' below then also name it 'having' here

routes/params.js
```diff
@@ -145,6 +145,13 @@ module.exports = {
     required: false,
     type: 'string',
   },
+  minimumPlayedParam: {
```
howardchung: and call it havingParam

wanderrful:
Right sorry -- I'm not familiar with this routing library that you're using here so I'm just sort of making it up as I go.  I've incorporated the changes you asked.

routes/spec.js
```diff
@@ -1447,6 +1448,7 @@ Please keep request rate to approximately 3/s.
             });
             const result = Object.keys(heroes)
               .map(k => heroes[k])
+              .filter(hero => !req.queryObj.having || hero.games_played >= Number(req.queryObj.having))
```
howardchung: looks like the field is "games" as the sort indicates below.  I had it wrong originally

howardchung: instead of games_played

wanderrful:
I've edited the "games_played" and made it into "games" for you, as you asked.  Hope this does the trick!

Issue #1575, opened 2017-12-07
Add API for scenarios data


howardchung:
A "scenario" is a particular combination of factors that are present in a match.

We want to store the count of matches where this happens, and the number of matches that were wins.

Examples
====
For the following: How often does this happen and what is the winrate when it does?
* Hero: Earthshaker is picked
* Hero + Item: Earthshaker buys a blink dagger at any point
* Hero + Item + Time: Earthshaker buys blink dagger before 10 minutes.  
* Hero + Stat: Earthshaker gets 10 kills
* Hero + Lane: Earthshaker goes mid
* Hero + Item + Tier: Earthshaker buys blink dagger in a Herald game

Design
====
* We would want to add a new postgres table called "scenarios" with the following columns:
  * hero
  * item
  * lh10 (and another column for each stat we might want to use in a scenario)
  * time (in seconds, as a before condition)
  * games
  * wins

After each parsed match, as part of the insertion step, evaluate the list of conditions (they will be "true" or "false" for each match and "won" or "lost") and increment the counts accordingly.
  * If "true" and "won", increment both games and wins
  * If "true" and "lost", increment games but not wins
  * Otherwise, ignore

We can assemble a list of conditions to check by permuting a list of items with cost over 2000 (as an example) and heroes, and adding any additional scenarios such as LH10 that we might care about.

Eventually we should be able to establish a list of conditions with high winrates, as well as allow users to query for scenarios involving certain heroes/items/etc. and discover what the winrates for those scenarios are.

Considerations:
We might need to do something to avoid duplicating inserts if the match is requested for reparse.  Possibly only do this if the source is scanner.

Since this step might be somewhat expensive, we may want to implement this logic as a new service, and have the parse insertion randomly push match IDs into a redis list.  The new service can consume this list as a queue, request the match object (using `buildMatch`) and do the scenario checks on that.

SheepD:
this sounds like a great epic.

some suggestions...

* i think keeping track of the games where the scenario isn't met is still a somewhat valuable metric. this would could give us an insight on how difficult it is to pull something like that off or ratio at which the scenario occurs in the wild.
* MMR might also be interesting to track in buckets(or mmr groups) so we can check if MMR plays a role in the success of a certain scenario

howardchung:
a "base" scenario consisting of just `hero_id` and no other columns could serve as a value to divide against.  That would always evaluate to true for a match that contained that hero.

[...]

PR #1585, opened 2018-01-03
[WIP] Scenarios

Argn0:
collecting data for #1575 

Second part of this will be building the APIs to read data from these tables.

Scenarios:

-Item timings for all heroes and all items that cost at least 2000 gold. Timings are  7.5, 10, 12, 15, 20, 25 and 30 minutes. e.g. Anti-Mage buys BF before minute 12.

-How often a player "pings". How does utilizing the ping feature impact a player's win rate? Does overusing this feature maybe lower team morale?

-The hero's lane role (safe, mid , offlane...). On which lane role does a hero have the best win rate?

-First Blood. What is a team's win rate if they get first blood in a match? How important is getting first blood?

-Courier Kill before minute 3.  What is a team's win rate if they get a courier kill before minute 3. How important is it to make sure your courier doesn't die before minute 3?

-Positivity or Negativity in chat. What is a team's win rate if they all chat positive or negative words before minute 1.  Does using positive words like "gl hf" boost team morale?

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -183,13 +183,12 @@
       "group": "socket",
       "exec_mode": "cluster",
       "instances": 1
-    }
-    ,
+    },
     {
       "script": "svc/scenarios.js",
       "watch": true,
       "ignore_watch": [".git", "node_modules"],
-      "group": "socket",
+      "group": "backend",
       "exec_mode": "cluster",
       "instances": 1
     }
--- a/manifest.json
+++ b/manifest.json
@@ -184,5 +184,14 @@
       "exec_mode": "cluster",
       "instances": 1
     }
+    ,
+    {
+      "script": "svc/scenarios.js",
+      "watch": true,
+      "ignore_watch": [".git", "node_modules"],
+      "group": "socket",
+      "exec_mode": "cluster",
+      "instances": 1
+    }
   ]
 }
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -421,12 +421,33 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
   rating int
 );
 
+CREATE TABLE IF NOT EXISTS scenarios (
+  scenario   TEXT,
+  hero       smallint,
+  item       TEXT,
+  time       smallint,
+  patch      smallint,
+  game_mode  smallint,
+  lobby_type smallint,
+  region     smallint,
+  games      BIGINT DEFAULT 0,
+  wins       BIGINT DEFAULT 0,
+  CONSTRAINT scenarios_constraint UNIQUE (hero, item, time, patch, game_mode, lobby_type, region)
+); 
+
 CREATE TABLE IF NOT EXISTS team_scenarios (
-  PRIMARY KEY (scenario),
-  scenario text,
-  wins bigint,
-  games bigint
-);
+  scenario   TEXT,
+  is_radiant BOOLEAN,
+  patch      smallint,
+  game_mode  smallint,
+  lobby_type smallint,
+  region     smallint,
+  games      BIGINT DEFAULT 0,
+  wins       BIGINT DEFAULT 0,
+  CONSTRAINT team_scenarios_constraint UNIQUE (is_radiant, patch, game_mode, lobby_type, region)
+);  
+ 
+
 
 DO $$
 BEGIN
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -428,8 +428,8 @@ CREATE TABLE IF NOT EXISTS scenarios (
   pings integer,
   game_duration smallint,
   lane smallint,
-  games      bigint DEFAULT 0,
-  wins       bigint DEFAULT 0,
+  games bigint DEFAULT 0,
+  wins bigint DEFAULT 0,
   CONSTRAINT scenario UNIQUE (hero, item, time),
   CONSTRAINT scenario2 UNIQUE (pings, game_duration),
   CONSTRAINT scenario3 UNIQUE (hero, lane, game_duration)
@@ -438,9 +438,9 @@ CREATE TABLE IF NOT EXISTS scenarios (
 CREATE TABLE IF NOT EXISTS team_scenarios (
   scenario   text,
   is_radiant BOOLEAN,
-  region     smallint,
-  games      bigint DEFAULT 0,
-  wins       bigint DEFAULT 0,
+  region smallint,
+  games bigint DEFAULT 0,
+  wins bigint DEFAULT 0,
   CONSTRAINT team_scenarios_constraint UNIQUE (scenario, is_radiant, region)
 );  
  
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -422,12 +422,16 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
 );
 
 CREATE TABLE IF NOT EXISTS scenarios (
-  hero       smallint,
-  item       text,
-  time       smallint,
+  scenario text DEFAULT '',
+  hero       smallint DEFAULT -1,
+  item       text DEFAULT '',
+  time       smallint DEFAULT -1,
+  pings integer DEFAULT -1,
+  game_duration smallint DEFAULT -1,
+  lane smallint DEFAULT -1,
   games      bigint DEFAULT 0,
   wins       bigint DEFAULT 0,
-  CONSTRAINT scenarios_constraint UNIQUE (hero, item, time)
+  CONSTRAINT scenarios_constraint UNIQUE (scenario, hero, item, time, pings, game_duration, lane)
 ); 
 
 CREATE TABLE IF NOT EXISTS team_scenarios (
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -421,6 +421,18 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
   rating int
 );
 
+CREATE TABLE IF NOT EXISTS team_scenarios (
+  PRIMARY KEY (team)
+  team boolean, -- 0 radiant / 1 dire
+  courier_kill boolean, -- achieved a courier kill before the 3 min mark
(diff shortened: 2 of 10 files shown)


manifest.json
```diff
@@ -184,5 +184,14 @@
       "exec_mode": "cluster",
       "instances": 1
     }
+    ,
```
howardchung: put the comma on the above line

manifest.json
```diff
@@ -184,5 +184,14 @@
...
+    {
+      "script": "svc/scenarios.js",
+      "watch": true,
+      "ignore_watch": [".git", "node_modules"],
+      "group": "socket",
```
howardchung: group should be backend

sql/create_tables.sql
```diff
@@ -421,6 +421,29 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
   rating int
 );
 
+CREATE TABLE IF NOT EXISTS scenarios (
+  hero smallint,
```
howardchung: suggest using hero_id for consistency with other tables

sql/create_tables.sql
```diff
@@ -421,6 +421,29 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
...
+  item text,
+  time smallint,
```
howardchung: suggest using int in case a game ends up going really long

sql/create_tables.sql
```diff
@@ -421,6 +421,29 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
...
+  pings integer,
+  game_duration smallint,
```
howardchung: use an int.

Is this field necessary?  Could we just use time instead?

sql/create_tables.sql
```diff
@@ -421,6 +421,29 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
...
+  games bigint DEFAULT 0,
+  wins bigint DEFAULT 0,
+  CONSTRAINT scenario UNIQUE (hero, item, time),
+  CONSTRAINT scenario2 UNIQUE (pings, game_duration),
+  CONSTRAINT scenario3 UNIQUE (hero, lane, game_duration)
+); 
+
+CREATE TABLE IF NOT EXISTS team_scenarios (
+  scenario   text,
+  is_radiant BOOLEAN,
```
howardchung: lowercase boolean

sql/create_tables.sql
```diff
@@ -421,6 +421,29 @@ CREATE TABLE IF NOT EXISTS leaderboard_rank (
...
+CREATE TABLE IF NOT EXISTS scenarios (
+  hero smallint,
+  item text,
+  time smallint,
+  pings integer,
+  game_duration smallint,
+  lane smallint,
+  games bigint DEFAULT 0,
+  wins bigint DEFAULT 0,
+  CONSTRAINT scenario UNIQUE (hero, item, time),
```
howardchung: I don't think these named constraints are necessary as long as your only insertion is via upsert on conflicting columns.

svc/parser.js
```diff
@@ -60,6 +61,11 @@ function parseProcessor(job, cb) {
       console.error(err.stack || err);
     } else {
       console.log('completed parse of match %s', match.match_id);
+      if (true || match.origin === 'scanner') {
+        if (Math.random() >= 0) {
```
howardchung: make this a config value in config.js

svc/parser.js
```diff
@@ -60,6 +61,11 @@ function parseProcessor(job, cb) {
       console.error(err.stack || err);
     } else {
       console.log('completed parse of match %s', match.match_id);
+      if (true || match.origin === 'scanner') {
```
howardchung: remove the `true`?

svc/parser.js
```diff
@@ -60,6 +61,11 @@ function parseProcessor(job, cb) {
...
+        if (Math.random() >= 0) {
+          redis.lpush('scenariosQueue', match.match_id);
```
howardchung: optional: add a monitor for this field to the buildStatus object so we can see if the queue is growing

svc/scenarios.js
```diff
@@ -0,0 +1,34 @@
...
+    if (err) {
+      cb(err);
+    }
+    su.scenarioChecks.forEach((scenarioCheck) => {
+      const rows = scenarioCheck(match);
+      async.eachSeries(rows, (row, cb) => {
+        const values = Object.keys(row.columns).map(() =>
+          '?');
+        const query = util.format(
+          `INSERT INTO %s (%s) VALUES (%s) ON CONFLICT (%s) DO UPDATE SET wins = ${row.table}.wins + ${row.won ? 1 : 0}, games = ${row.table}.games + 1`,
```
howardchung: you can use EXCLUDED.wins to avoid the need to interpolate a raw string into the query

howardchung: I suggest parameterizing the row.won ? 1 : 0 part with a `?` to lower risk of sql injection

howardchung: I think you need to set the initial values here for wins and games, otherwise your counts will be off by 1 (the first insert will result in 0 for both fields)

util/scenariosUtil.js
```diff
@@ -0,0 +1,143 @@
...
+const scenarioChecks = [
+
+  function itemTimings(match) {
+    const rows = [];
+    if (match.players) {
+      match.players.forEach((player) => {
+        if (player.purchase_log) {
+          player.purchase_log.forEach((item) => {
+            if (dotaItems.indexOf(item.key) !== -1 && item.time <= timings[timings.length - 1]) {
+              const won = (player.player_slot < 5 && match.radiant_win) || (player.player_slot > 4 && !match.radiant_win);
```
howardchung: There is a utility function called isRadiant that you can use here

util/scenariosUtil.js
```diff
@@ -0,0 +1,143 @@
...
+              rows.push({
+                columns: {
+                  hero: player.hero_id,
+                  item: item.key,
+                  time: timings.find(x => x >= item.time),
```
howardchung: you can add the wins column here as a field instead of a separate prop called won.  It will be 1 or 0n and the upsert logic can handle it automatically

util/scenariosUtil.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    return rows;
+  },
+
+  function pings(match) { // how often a player "pings"
+    const rows = [];
+    if (match.players) {
+      match.players.forEach((player) => {
+        if (player.pings <= pingBucket[pingBucket.length - 1]) {
+          const pings = pingBucket.find(x => x >= player.pings) || -2;
```
howardchung: what is -2?

util/scenariosUtil.js
```diff
@@ -0,0 +1,143 @@
...
+          const won = (player.player_slot < 5 && match.radiant_win) || (player.player_slot > 4 && !match.radiant_win);
```
howardchung: use isRadiant util function

util/scenariosUtil.js
```diff
@@ -0,0 +1,143 @@
...
+    let radiantCondition = false;
+    let direCondition = false;
+    if (match.chat) {
+      for (let i = 0; i < match.chat.length; i += 1) {
+        const c = match.chat[i];
+        if (c.time >= 600) {
+          break;
+        }
+        if (negativeWords.some(word => c.key.toLowerCase().indexOf(word) !== -1)) {
+          if (c.slot < 5) {
```
howardchung: use the `player_slot` field, then you can also use the utility function here

Argn0: I think that won't work because the utility function takes a player object and the the chat log is in match.chat

Argn0:
-constraints are no longer named
-first insert now works correctly and no longer results in 0 for both fields
-all query values are now parameterized
TODO: add more scenarios

sql/migrations.sql
```diff
@@ -10,3 +10,25 @@ CREATE INDEX IF NOT EXISTS public_matches_avg_rank_tier_idx on public_matches(av
 ALTER TABLE player_matches ADD ability_targets json;
 
 ALTER TABLE public_matches ADD cluster integer;
+
+CREATE TABLE IF NOT EXISTS scenarios (
+  hero_id smallint,
+  item text,
+  time integer,
+  pings integer,
+  lane smallint,
```
howardchung: we may want to use lane_role instead of lane, since lane is absolute position (top/mid/bot) and lane_role is normalized for side (safe/mid/off)

svc/scenarios.js
```diff
@@ -0,0 +1,37 @@
...
+        const rows = scenarioCheck(match);
+        async.eachSeries(rows, (row, cb) => {
+          const values = Object.keys(row).map(() =>
+            '?');
+          const query = util.format(
+            'INSERT INTO %s (%s) VALUES (%s) ON CONFLICT (%s) DO UPDATE SET wins = (%s).wins + EXCLUDED.wins, games = (%s).games + 1',
+            table,
+            Object.keys(row).join(','),
+            values,
+            Object.keys(row).slice(0, -1).join(','),
```
howardchung: what's the purpose of .slice(0, -1) here?

svc/scenarios.js
```diff
@@ -0,0 +1,37 @@
...
+      cb(err);
+    }
+    Object.keys(su.scenarioChecks).forEach((table) => {
+      su.scenarioChecks[table].forEach((scenarioCheck) => {
+        const rows = scenarioCheck(match);
+        async.eachSeries(rows, (row, cb) => {
+          const values = Object.keys(row).map(() =>
+            '?');
+          const query = util.format(
+            'INSERT INTO %s (%s) VALUES (%s) ON CONFLICT (%s) DO UPDATE SET wins = (%s).wins + EXCLUDED.wins, games = (%s).games + 1',
```
howardchung: I don't think you need the parentheses for (%s).wins and (%s).games

Argn0:
-replaced slice method with filter method. (wins column is not an ON CONFLICT target)
-removed unnecessary  parentheses in query string
-replaced "lane" with "lane_role"

Issue #1598, opened 2018-01-31
exception in scenarios worker


howardchung:
Getting the following exception occasionally:

@Argn0 
```
17|scenari | TypeError: Cannot read property 'forEach' of null
17|scenari |     at match.players.forEach (/usr/src/util/scenariosUtil.js:32:29)
17|scenari |     at Array.forEach (<anonymous>)
17|scenari |     at itemTimings (/usr/src/util/scenariosUtil.js:31:21)
17|scenari |     at su.scenarioChecks.(anonymous function).forEach (/usr/src/svc/scenarios.js:15:22)
17|scenari |     at Array.forEach (<anonymous>)
17|scenari |     at Object.keys.forEach (/usr/src/svc/scenarios.js:14:32)
17|scenari |     at Array.forEach (<anonymous>)
17|scenari |     at buildMatch (/usr/src/svc/scenarios.js:13:36)
17|scenari |     at Command.redis.setex [as callback] (/usr/src/store/buildMatch.js:178:18)
17|scenari |     at normal_reply (/usr/src/node_modules/redis/index.js:726:21)
17|scenari |     at RedisClient.return_reply (/usr/src/node_modules/redis/index.js:824:9)
17|scenari |     at JavascriptRedisParser.returnReply (/usr/src/node_modules/redis/index.js:192:18)
17|scenari |     at JavascriptRedisParser.execute (/usr/src/node_modules/redis-parser/lib/parser.js:574:12)
17|scenari |     at Socket.<anonymous> (/usr/src/node_modules/redis/index.js:274:27)
17|scenari |     at emitOne (events.js:116:13)
17|scenari |     at Socket.emit (events.js:211:7)
```

PR #1599, opened 2018-01-31
Fix for exception in scenarios worker. Match.players is null

Argn0:
fixes https://github.com/odota/core/issues/1598

### DIFF
--- a/svc/scenarios.js
+++ b/svc/scenarios.js
@@ -34,7 +34,7 @@ function processScenarios(matchID, cb) {
         });
       });
     });
-    cb();
+    return cb();
   });
 }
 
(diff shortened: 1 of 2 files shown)


svc/scenarios.js
```diff
@@ -10,6 +10,10 @@ function processScenarios(matchID, cb) {
     if (err) {
       cb(err);
     }
+    if (!su.validateMatchProperties) {
+      console.log(`Skipping scenario checks for match ${matchID}. Invalid match object.`);
```
howardchung: use console.error

Argn0:
I don't know why that test randomly broke. It passes on my machine.

svc/scenarios.js
```diff
@@ -10,6 +10,10 @@ function processScenarios(matchID, cb) {
     if (err) {
       cb(err);
```
howardchung: you actually need to return `cb(err)` in both places to prevent execution of the rest of the code.

That's probably why we encountered the invalid matches in the first place, actually.

PR #1605, opened 2018-02-09
0 MMR value fix

Argn0:


howardchung:
note that this won't fix the values already in the database so the client side fix is still needed.

PR #1606, opened 2018-02-14
Update processDraftTimings.js

bobsaget4lyfe:
Need to test if it is calculating the timings correctly. This should fix the active team though. The problem is that the parser grabs when a pick or ban has happened so that active team is one step too late. So the actual active team is from the previous pick/ban. For the first pick/ban need to sum all the picks and bans except the final one and modulus by 2 then add by 2 to find the relevant active team.

### DIFF
--- a/processors/processDraftTimings.js
+++ b/processors/processDraftTimings.js
@@ -16,16 +16,18 @@
 function processDraftTimings(entries, meta) {
   const draftTimings = [];
   const heroIdToSlot = meta.hero_id_to_slot;
+  const sumActiveTeam = 0;
   for (let i = 0; i < entries.length; i += 1) {
     const e = entries[i];
     const heroId = e.hero_id;
     if (e.type === 'draft_timings') {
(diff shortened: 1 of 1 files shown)


processors/processDraftTimings.js
```diff
@@ -16,16 +16,18 @@
 function processDraftTimings(entries, meta) {
   const draftTimings = [];
   const heroIdToSlot = meta.hero_id_to_slot;
+  const sumActiveTeam = 0;
   for (let i = 0; i < entries.length; i += 1) {
     const e = entries[i];
     const heroId = e.hero_id;
     if (e.type === 'draft_timings') {
+      sumActiveTeam = i < entreies.length - 1 ? sumActiveTeam + e.draft_active_team : sumActiveTeam;
```
howardchung: entries

howardchung: if you are reassigning sumActiveTeam you need to use let instead of const

PR #1610, opened 2018-02-17
Scenarios maintenance service

Argn0:


package-lock.json
```diff
@@ -259,11 +259,6 @@
       "resolved": "https://registry.npmjs.org/async-each/-/async-each-1.0.1.tgz",
```
howardchung: I don't think this file should change

svc/scheduler.js
```diff
@@ -0,0 +1,95 @@
...
+const db = require('../store/db');
+const async = require('async');
+const fs = require('fs');
+const config = require('../config');
+const apps = require('../manifest.json').apps;
+
+
+const scenariosSvc = apps.find(x => x.script === 'svc/scenarios.js');
+
+function getStatus(process, cb) {
```
howardchung: This approach won't work in prod since the filesystem and pm2 uptime are ephemeral (they restart at least once a day).

If you want to track state, you'll need to store it in either redis or postgres and then read it in your service.

Argn0:
how about we use the table "scenarios" in production and collect data in a table "scenarios_new". when "scenarios_new" has enough data we rename "scenarios" to "scenarios_old". rename "scenarios_new" to "scenarios". then drop table "scenarios_old".

Argn0:
Or we tag new rows somehow. When we are finished collecting data we remove all rows that do not have this tag. Then we remove the tags from the rows. I think I like this idea better.

Argn0:
I added a new boolean column "current" that defaults to TRUE.

5 days before the table gets reset we start collecting data. When the table gets reset it deletes all rows which have a current column that is NOT TRUE. Then it changes every row's current column to FALSE. We stop collecting data until the next cycle.

sql/create_tables.sql
```diff
@@ -430,9 +430,10 @@ CREATE TABLE IF NOT EXISTS scenarios (
   lane_role smallint,
   games bigint DEFAULT 1,
   wins bigint,
-  UNIQUE (hero_id, item, time),
-  UNIQUE (pings, time),
-  UNIQUE (hero_id, lane_role, time)
+  current boolean DEFAULT true,
```
howardchung: I suggest making this a `timestamp with time zone` column and set it to the beginning of the period that you want to track (week, month, etc.)

Then you can just do a range delete query to clean up the data.

howardchung: moment.js library has handy helpers to fetch startOf('week') and startOf('month')

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,107 @@
...
+      if (err) {
+        cb(err);
+      }
+      const today = new Date();
+      const scenariosReset = reply;
+      const dayDifference = Math.ceil(Math.abs(today.getTime() - new Date(scenariosReset).getTime()) / (1000 * 3600 * 24));
+      if (dayDifference >= config.SCENARIOS_TABLES_RESET_PERIOD) {
+        async.parallel([
+          (cb) => {
+            db('team_scenarios').where('current', '!=', 'true').del().then(() => db('team_scenarios').update({
```
howardchung: with the timestamp approach you can probably just do a single `DELETE FROM team_scenarios WHERE timestamp < now() - interval '30 day'` and repeat for scenarios

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,107 @@
...
+          redis.set('scenarios_reset_date', new Date());
+          cb();
+        });
+      } else {
+        cb();
+      }
+    });
+  },
+  // collect new data before old data gets deleted
+  function scheduleScenariosSvc(cb) {
```
howardchung: I don't think this is needed if you use the timestamp approach, the service can just run continuously

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,107 @@
...
+
+redis.get('scenarios_reset_date', (err, reply) => {
+  if (err) {
+    console.error(err);
+  } else if (!reply) {
+    redis.set('scenarios_reset_date', new Date());
+  }
+});
+
+function start() {
```
howardchung: I think you can just use the invokeInterval method used by several of the other services to run your function every x seconds

sql/migrations.sql
```diff
@@ -32,3 +32,17 @@ CREATE TABLE IF NOT EXISTS team_scenarios (
...
+ALTER TABLE team_scenarios ADD epoch_week smallint;
+ALTER TABLE scenarios DROP CONSTRAINT "scenarios_hero_id_item_time_key";
+ALTER TABLE scenarios DROP CONSTRAINT "scenarios_hero_id_lane_role_time_key";
+ALTER TABLE scenarios DROP CONSTRAINT "scenarios_pings_time_key";
+ALTER TABLE team_scenarios DROP CONSTRAINT "team_scenarios_scenario_is_radiant_region_key";
+ALTER TABLE scenarios ADD UNIQUE (hero_id, item, time, epoch_week);
+ALTER TABLE scenarios ADD UNIQUE (pings, time, epoch_week);
+ALTER TABLE scenarios ADD UNIQUE (hero_id, lane_role, time, epoch_week);
+ALTER TABLE team_scenarios ADD UNIQUE (scenario, is_radiant, region, epoch_week);
+--- !!!
```
howardchung: what is this line for?

sql/migrations.sql
```diff
@@ -32,3 +32,17 @@ CREATE TABLE IF NOT EXISTS team_scenarios (
...
+DELETE FROM team_scenarios WHERE epoch_week IS NULL;
```
howardchung: I would suggest just building this delete into the service (i.e. delete rows without epoch_week)

svc/scenarios.js
```diff
@@ -14,17 +14,19 @@ function processScenarios(matchID, cb) {
...
+    const currentWeek = Math.floor(new Date() / (1000 * 60 * 60 * 24 * 7));
     Object.keys(su.scenarioChecks).forEach((table) => {
       su.scenarioChecks[table].forEach((scenarioCheck) => {
         const rows = scenarioCheck(match);
         async.eachSeries(rows, (row, cb) => {
           const values = Object.keys(row).map(() =>
             '?');
           const query = util.format(
-            'INSERT INTO %s (%s) VALUES (%s) ON CONFLICT (%s) DO UPDATE SET wins = %s.wins + EXCLUDED.wins, games = %s.games + 1',
+            'INSERT INTO %s (%s, epoch_week) VALUES (%s, %s) ON CONFLICT (%s, epoch_week) DO UPDATE SET wins = %s.wins + EXCLUDED.wins, games = %s.games + 1',
```
howardchung: I would suggest including epoch_week into the row object to simplify your query

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,47 @@
+const db = require('../store/db');
+const async = require('async');
+const config = require('../config');
+
```
howardchung: extra line

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,47 @@
...
+      if (err) {
+        cb(err);
+      }
+      cb();
+    });
+  },
+];
+
+setInterval(() => {
+  async.eachSeries(tasks, (task, cb) => {
```
howardchung: it looks like you have only one task in the tasks array, is this async.eachSeries necessary?

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,47 @@
+const db = require('../store/db');
+const async = require('async');
+const config = require('../config');
+
+
+const tasks = [
+  function clearScenariosTables(cb) {
+    const currentWeek = Math.floor(new Date() / (1000 * 60 * 60 * 24 * 7));
```
howardchung: you may want to move this to utility file since you're using it both here and in insert

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,47 @@
...
+    async.parallel([
+      (cb) => {
+        db('team_scenarios').where('epoch_week', '<=', currentWeek - config.MAXIMUM_AGE_SCENARIOS_ROWS).del()
+          .asCallback((err) => {
```
howardchung: you can shorthand this to just .asCallback(cb)

svc/scenariosCleanup.js
```diff
@@ -0,0 +1,47 @@
...
+        db('scenarios').where('epoch_week', '<=', currentWeek - config.MAXIMUM_AGE_SCENARIOS_ROWS).del()
+          .asCallback((err) => {
+            if (err) {
+              cb(err);
+            } else {
+              cb();
+            }
+          });
+      },
+    ], (err) => {
```
howardchung: just `cb`

sql/create_tables.sql
```diff
@@ -430,9 +430,10 @@ CREATE TABLE IF NOT EXISTS scenarios (
   lane_role smallint,
   games bigint DEFAULT 1,
   wins bigint,
-  UNIQUE (hero_id, item, time),
-  UNIQUE (pings, time),
-  UNIQUE (hero_id, lane_role, time)
+  epoch_week smallint,
```
howardchung: suggest using the int or bigint types, the bytes saved are probably not that significant and we don't want to run out of space

config.js
```diff
@@ -57,6 +57,7 @@ const defaults = {
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
   WEBSOCKET_PORT: 80, // port for live match subscription websocket service
   ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication
+  MAXIMUM_AGE_SCENARIOS_ROWS: 3, // maximum allowed age of scenarios rows in weeks
```
howardchung: If you only need current week and previous week, I think this can be 2, you can use < in your postgres query instead of <=

sql/create_tables.sql
```diff
@@ -430,9 +430,10 @@ CREATE TABLE IF NOT EXISTS scenarios (
   lane_role smallint,
   games bigint DEFAULT 1,
   wins bigint,
-  UNIQUE (hero_id, item, time),
-  UNIQUE (pings, time),
-  UNIQUE (hero_id, lane_role, time)
+  epoch_week integer,
+  UNIQUE (hero_id, item, time, epoch_week),
```
howardchung: I'm not sure if these constraints are needed anymore since you're not referencing them by name anymore.  They might be useful if you want to ensure that you don't accidentally have a duplicate, but if your scenario insert logic is correct that should never happen.

PR #1612, opened 2018-02-18
Don't include pro in list of own matches with pro players

### LINKED ISSUE #1607: Bug in https://api.opendota.com/api/players/{account_id}/pros
When calling this endpoint for a notable player, the response array contains said notable player as well.

njgingrich:
Fixes #1607.

### DIFF
--- a/store/queries.js
+++ b/store/queries.js
@@ -367,9 +367,10 @@ function getProPeers(db, input, player, cb) {
       return cb(err);
     }
     const arr = result.rows.map(r =>
-      Object.assign({}, r, teammates[r.account_id])).filter(r =>
-      r.games).sort((a, b) =>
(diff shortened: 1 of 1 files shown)


store/queries.js
```diff
@@ -367,9 +367,10 @@ function getProPeers(db, input, player, cb) {
       return cb(err);
     }
     const arr = result.rows.map(r =>
-      Object.assign({}, r, teammates[r.account_id])).filter(r =>
-      r.games).sort((a, b) =>
-      b.games - a.games);
+      Object.assign({}, r, teammates[r.account_id]))
+      .filter(r => r.account_id !== player.account_id)
```
howardchung: you can combine both filters into one .filter function I think

PR #1621, opened 2018-03-14
Scenarios endpoints

Argn0:


routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
         },
       },
     },
+    '/scenarios/{hero_id}/itemTimings': {
+      get: {
+        summary: 'GET /scenarios/{hero_id}/itemTimings',
+        description: 'Win rates for certain item timings',
```
howardchung: "on a hero".

You may want to consider flipping the param order to /itemTimings/hero_id if you want to support querying all item timings (not providing a hero_id) returns general item timings win rates

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+        tags: ['scenarios'],
+        parameters: [{
+          name: 'item',
+          in: 'query',
+          description: 'Filter by item name',
```
howardchung: may want to provide an example item name

routes/spec.js
howardchung: You may want to consider making both hero_id and item query params, so the user can filter on them independently.  e.g. a user can request win rates for blink dagger on all heroes, or win rates for all items on Riki

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                  item: {
+                    description: 'Purchased item',
+                    type: 'string',
+                  },
+                  time: {
+                    description: 'Ingame time in seconds before the item was purchased',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'In how many games was the item bought by this hero within that time',
```
howardchung: "the number of games where the hero bought this item before this time"

howardchung: actually i forget, are the item timings rounded down to the nearest 5 minute block?

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                    type: 'string',
+                  },
+                  wins: {
+                    description: 'How many games were won when this hero bought the item within that time',
```
howardchung: the number of games won where the hero bought this item before this time

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                    description: 'Purchased item',
+                    type: 'string',
+                  },
+                  time: {
+                    description: 'Ingame time in seconds before the item was purchased',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'In how many games was the item bought by this hero within that time',
+                    type: 'string',
```
howardchung: are these integers?

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                    description: 'How many games were won when this hero bought the item within that time',
+                    type: 'string',
+                  },
+                },
+              },
+            },
+          },
+        },
+        route: () => '/scenarios/:hero_id/itemTimings',
+        func: (req, res, cb) => {
```
howardchung: consider moving this function to queries.js, the eventual goal is to put all of them there (I haven't gotten around to some of them

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+          const heroId = req.params.hero_id;
+          const item = req.query.item;
+
+          if (item) {
+            db.raw(`SELECT item, time, sum(games) games, sum(wins) wins
```
howardchung: you should be able to use a single query template for this by doing something like `AND (? IS NULL or item = ?)` (basically make the first condition true when no item/empty string/null is passed)

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                if (err) {
+                  return cb(err);
+                }
+                return res.json(result.rows);
+              });
+          }
+        },
+      },
+    },
+    '/scenarios/{hero_id}/laneRoles': {
```
howardchung: same comment about maybe inverting the order

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                    description: 'The hero\'s lane role',
+                    type: 'integer',
+                  },
+                  time: {
+                    description: 'Total up to game length in seconds',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'In how many games this hero was played in this lane role',
+                    type: 'string',
```
howardchung: integer?

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+              type: 'array',
+              items: {
+                type: 'object',
+                properties: {
+                  lane_role: {
+                    description: 'The hero\'s lane role',
+                    type: 'integer',
+                  },
+                  time: {
+                    description: 'Total up to game length in seconds',
```
howardchung: is this the full game length?

routes/spec.js
```diff
@@ -3943,6 +3944,186 @@ Please keep request rate to approximately 3/s.
...
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'In how many games this hero was played in this lane role',
```
howardchung: same comment about phrasing on this scenario

util/scenariosUtil.js
```diff
@@ -148,8 +148,15 @@ function validateMatchProperties(match) {
   return matchProperties.every(property => match[property] !== undefined && match[property] !== null);
 }
 
+const teamScenariosQueryParams = {
+  pos_chat_1min: 'Positivity in chat before 1min',
```
howardchung: expand this to "1 minute"

routes/spec.js
Argn0: they are  [7.5, 10, 12, 15, 20, 25, 30]

so 0 - 7.5 min  ,  7.5 - 10 min , 10 - 12 min  etc.

routes/spec.js
Argn0: no they are strings in the JSON response

Argn0:
Moved all functions to queries.js and fixed everything else you mentioned.

routes/params.js
```diff
@@ -201,4 +201,11 @@ module.exports = {
     required: false,
     type: 'integer',
   },
+  heroIdParamPath: {
+    name: 'hero_id',
+    in: 'path',
```
howardchung: this is now a query param

routes/spec.js
```diff
@@ -3943,6 +3944,178 @@ Please keep request rate to approximately 3/s.
...
+                  lane_role: {
+                    description: 'The hero\'s lane role',
+                    type: 'integer',
+                  },
+                  time: {
+                    description: 'Maximum game length in seconds',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'The number of games won where the hero bought this item before this time',
```
howardchung: i think this the description for the field below,

This one should be just number of games

routes/spec.js
```diff
@@ -3943,6 +3944,178 @@ Please keep request rate to approximately 3/s.
...
+                  item: {
+                    description: 'Purchased item',
+                    type: 'string',
+                  },
+                  time: {
+                    description: 'Ingame time in seconds before the item was purchased',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'The number of games won where the hero bought this item before this time',
```
howardchung: this should be for the field below

store/queries.js
```diff
@@ -1263,6 +1264,40 @@ function insertMatch(match, options, cb) {
   });
 }
 
+function getItemTimings(req, cb) {
+  const heroId = req.query.hero_id || 0;
+  const item = req.query.item || 0;
+  db.raw(
+    `SELECT hero_id, item, time, sum(games) games, sum(wins) wins FROM scenarios WHERE item IS NOT NULL
+       AND (0 = :heroId OR hero_id = :heroId) AND ('0' = :item OR item = :item)
```
howardchung: does '0' comparison work?  shouldn't it be 0 since it is not a string?

howardchung: Would advise splitting each WHERE/AND on its own line

store/queries.js
```diff
@@ -1263,6 +1264,40 @@ function insertMatch(match, options, cb) {
...
+  )
+    .asCallback((err, result) => cb(err, result));
+}
+
+function getLaneRoles(req, cb) {
+  const heroId = req.query.hero_id || 0;
+  const lane = req.query.lane_role || 0;
+  db.raw(
+    `SELECT hero_id, lane_role, time, sum(games) games, sum(wins) wins 
+       FROM scenarios WHERE lane_role IS NOT NULL
```
howardchung: same formatting of SQL

store/queries.js
```diff
@@ -1263,6 +1264,40 @@ function insertMatch(match, options, cb) {
...
+    { heroId, lane },
+  )
+    .asCallback((err, result) => cb(err, result));
+}
+
+function getTeamScenarios(req, cb) {
+  const scenario = su.teamScenariosQueryParams[req.query.scenario] || 0;
+
+  db.raw(`SELECT scenario, is_radiant, region, sum(games) games, sum(wins) wins 
+    from team_scenarios WHERE (0 = :scenario OR scenario = :scenario)
```
howardchung: format the query

util/scenariosUtil.js
```diff
@@ -121,16 +122,16 @@ const scenarioChecks = {
           }
         }
         if (radiantNegative) {
-          rows.push(buildTeamScenario('Negativity in chat before 1min', true, match)[0]);
+          rows.push(buildTeamScenario('Negativity in chat before 1 minute', true, match)[0]);
```
howardchung: you can probably reuse the teamScenariosQueryParams strings here

Argn0:
- formatted all query strings
- fixed games/wins response descriptions
- removed heroIdParamPath from params.js I was not using it anymore

routes/spec.js
```diff
@@ -3943,6 +3944,178 @@ Please keep request rate to approximately 3/s.
...
+              type: 'array',
+              items: {
+                type: 'object',
+                properties: {
+                  scenario: {
+                    description: 'The scenario\'s name or description',
+                    type: 'string',
+                  },
+                  is_radiant: {
+                    description: 'Radiant or Dire',
```
howardchung: "boolean indicating if the player was radiant" (i feel like we have this property a few times already, it should probably be refactored into one place)

routes/spec.js
```diff
@@ -3943,6 +3944,178 @@ Please keep request rate to approximately 3/s.
...
+                    type: 'boolean',
+                  },
+                  region: {
+                    decription: 'Region the game was played in',
+                    type: 'integer',
+                  },
+                  games: {
+                    description: 'The number of games where this scenario occured',
```
howardchung: occurred

store/queries.js
```diff
@@ -1263,6 +1264,48 @@ function insertMatch(match, options, cb) {
...
+  db.raw(
+    `SELECT hero_id, item, time, sum(games) games, sum(wins) wins
+     FROM scenarios
+     WHERE item IS NOT NULL
+     AND (0 = :heroId OR hero_id = :heroId)
+     AND ('' = :item OR item = :item)
+     GROUP BY hero_id, item, time ORDER BY time, hero_id, item
+     LIMIT 1600`,
+    { heroId, item },
+  ).asCallback((err, result) => cb(err, result));
```
howardchung: you can actually shorthand this to `asCallback(cb)`, but this is fine too, it makes it easier to add logging if we need to

routes/spec.js
```diff
@@ -238,7 +239,7 @@ Please keep request rate to approximately 3/s.
                   type: 'integer',
                 },
                 radiant_win: {
-                  description: 'Boolean indicating whether Radiant won the match',
+                  description: radiantWinDescr,
```
howardchung: probably the entire radiant_win object can be refactored out

howardchung: you might actually want to create a new file in https://github.com/odota/core/tree/master/routes called properties.js and put commonly used properties in there, then import them into this file

routes/spec.js
```diff
@@ -4085,15 +4086,15 @@ Please keep request rate to approximately 3/s.
                     type: 'string',
                   },
                   is_radiant: {
-                    description: 'Radiant or Dire',
+                    description: radiantWinDescr,
```
howardchung: this one is not the same as the others

PR #1622, opened 2018-03-18
[WIP] Meter API usage via API keys

albertcui:


config.js
```diff
@@ -56,11 +56,13 @@ const defaults = {
   ENABLE_PLAYER_CACHE: 1, // enable/disable player aggregation caching
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
   WEBSOCKET_PORT: 80, // port for live match subscription websocket service
-  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication
   MAXIMUM_AGE_SCENARIOS_ROWS: 2, // maximum allowed age of scenarios rows in weeks
   MATCH_CACHE_SECONDS: 60, // number of seconds to cache matches
   PLAYER_CACHE_SECONDS: 3600, // number of seconds to cache player aggregations
   SCANNER_PLAYER_PERCENT: 10, // percent of matches from scanner to insert player account IDs for (discover new player account IDs)
+  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication,
+  API_FREE_LIMIT: 25000, // number of api requests per month before 429 is returned. If using an API key, calls over this are charged.
```
howardchung: we may want to set this to 0 (disabled) here and use config to set it on GCP since users running their own copies probably don't need this.

config.js
```diff
@@ -56,11 +56,13 @@ const defaults = {
...
+  API_KEY_PER_MIN_LIMIT: 180, // Rate limit per minute if using an API key
```
howardchung: add a config for rate limit without API key?

sql/create_tables.sql
```diff
@@ -182,6 +182,27 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
 
+CREATE TABLE IF NOT EXISTS api_keys (
+  PRIMARY KEY(account_id),
+  account_id bigint UNIQUE,
+  api_key uuid UNIQUE,
+  customer_id varchar(255)
+);
+CREATE INDEX IF NOT EXISTS api_keys_account_id_idx on api_keys(account_id);
+CREATE INDEX IF NOT EXISTS api_keys_customer_id_idx on api_keys(customer_id);
+CREATE INDEX IF NOT EXISTS api_keys_keys on api_keys(api_key);
```
howardchung: this should be named api_keys_api_key

sql/create_tables.sql
```diff
@@ -182,6 +182,27 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
 CREATE INDEX IF NOT EXISTS subscriptions_customer_id_idx on subscriptions(customer_id);
 
+CREATE TABLE IF NOT EXISTS api_keys (
+  PRIMARY KEY(account_id),
+  account_id bigint UNIQUE,
+  api_key uuid UNIQUE,
+  customer_id varchar(255)
+);
+CREATE INDEX IF NOT EXISTS api_keys_account_id_idx on api_keys(account_id);
+CREATE INDEX IF NOT EXISTS api_keys_customer_id_idx on api_keys(customer_id);
```
howardchung: what is customer_id used for?

sql/create_tables.sql
```diff
@@ -182,6 +182,27 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+CREATE INDEX IF NOT EXISTS api_keys_keys on api_keys(api_key);
+
+CREATE TABLE IF NOT EXISTS api_key_usage (
+  account_id bigint REFERENCES api_keys(account_id),
```
howardchung: may want ON DELETE CASCADE here

sql/create_tables.sql
```diff
@@ -182,6 +182,27 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+  customer_id varchar(255),
+  api_key uuid,
+  usage_count int,
```
howardchung: bigint in case it goes over 2 billion?

sql/create_tables.sql
```diff
@@ -182,6 +182,27 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+  PRIMARY KEY(account_id),
+  account_id bigint UNIQUE,
+  api_key uuid UNIQUE,
+  customer_id varchar(255)
+);
+CREATE INDEX IF NOT EXISTS api_keys_account_id_idx on api_keys(account_id);
+CREATE INDEX IF NOT EXISTS api_keys_customer_id_idx on api_keys(customer_id);
+CREATE INDEX IF NOT EXISTS api_keys_keys on api_keys(api_key);
+
+CREATE TABLE IF NOT EXISTS api_key_usage (
```
howardchung: what is this table used for?  aren't we tracking usage in redis?

svc/web.js
```diff
@@ -17,11 +17,28 @@ const passport = require('passport');
...
+
+setTimeout(() => {
+  queries.getAPIKeys(db, (err, row) => {
+    if (err) {
+      return;
+    }
+
+    OD_API_KEYS = {};
+
+    row.forEach((e) => {
```
howardchung: keep in mind this will block the event loop and will not scale very far.  The web server cannot serve requests while it is in this loop.

svc/web.js
```diff
@@ -17,11 +17,28 @@ const passport = require('passport');
...
+      OD_API_KEYS[e.api_key] = 0;
```
howardchung: why 0 and not 1?

svc/web.js
```diff
@@ -17,11 +17,28 @@ const passport = require('passport');
...
+const { redisCount } = utility;
 
 const app = express();
 const apiKey = config.STEAM_API_KEY.split(',')[0];
 const host = config.ROOT_URL;
+
+let OD_API_KEYS = {};
+
+setTimeout(() => {
+  queries.getAPIKeys(db, (err, row) => {
```
howardchung: rows?

svc/web.js
```diff
@@ -68,27 +85,56 @@ app.use('/ugc', (req, res) => {
...
-  console.log('%s visit %s, ip %s', req.user ? req.user.account_id : 'anonymous', req.originalUrl, ip);
-  redis.multi().incr(key).expireat(key, utility.getStartOfBlockMinutes(1, 1)).exec((err, resp) => {
-    if (err) {
-      console.log(err);
+  [ip] = ip.replace(/^.*:/, '').split(',');
+
+  let rateKey = '';
+  let countKey = '';
+
+  const requestAPIKey = req.query.OPENDOTA_API_KEY;
```
howardchung: shouldn't this just be `key` and not OPENDOTA_API_KEY

svc/web.js
```diff
@@ -68,27 +85,56 @@ app.use('/ugc', (req, res) => {
...
+    .exec((err, resp) => {
+      if (err) {
+        console.log(err);
+        return cb();
+      }
+      if (config.NODE_ENV === 'development') {
+        console.log(resp);
+      }
+
+      if (((isAPIRequest && resp[0] > config.API_KEY_PER_MIN_LIMIT) || resp[0] > 60)
```
howardchung: you can determine the limit to use in a line above and then only make one comparison here

svc/web.js
```diff
@@ -68,27 +85,56 @@ app.use('/ugc', (req, res) => {
...
+          && config.NODE_ENV !== 'test') {
+        return res.status(429).json({
+          error: 'rate limit exceeded',
+        });
+      }
+
+      if (!isAPIRequest && resp[2] > config.API_FREE_LIMIT && config.NODE_ENV !== 'test') {
+        return res.status(429).json({
+          error: 'monthly api limit exeeded',
```
howardchung: I feel that we should swap this to daily or weekly, blocking an IP for up to a full month sounds potentially bad

sql/create_tables.sql
albertcui: billing

svc/web.js
albertcui: I wanted to name space it. Wasn't sure if we were using key :P

svc/web.js
albertcui: It makes charging too complicated.

svc/web.js
albertcui: What do we do instead? :P

sql/create_tables.sql
albertcui: This table is the source of truth. We can track usage over time this way.

svc/web.js
albertcui: IDK I'll change it :P

config.js
```diff
@@ -56,11 +56,15 @@ const defaults = {
...
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
   WEBSOCKET_PORT: 80, // port for live match subscription websocket service
-  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication
   MAXIMUM_AGE_SCENARIOS_ROWS: 2, // maximum allowed age of scenarios rows in weeks
   MATCH_CACHE_SECONDS: 60, // number of seconds to cache matches
   PLAYER_CACHE_SECONDS: 3600, // number of seconds to cache player aggregations
   SCANNER_PLAYER_PERCENT: 10, // percent of matches from scanner to insert player account IDs for (discover new player account IDs)
+  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication,
+  ENABLE_API_LIMIT: '',
+  API_FREE_LIMIT: 25000, // number of api requests per month before 429 is returned. If using an API key, calls over this are charged.
```
howardchung: disable this limitation by default

howardchung: looks like you create ENABLE_API_LIMIT but why not just use a 0 value to reduce the amount of config?

sql/create_tables.sql
```diff
@@ -379,6 +399,7 @@ CREATE TABLE IF NOT EXISTS hero_ranking (
   score double precision
 );
 CREATE INDEX IF NOT EXISTS hero_ranking_hero_id_score_idx ON hero_ranking(hero_id, score);
+CREATE INDEX IF NOT EXISTS hero_ranking_score_idx ON hero_ranking(score);
```
howardchung: remove this

sql/create_tables.sql
```diff
@@ -160,7 +159,7 @@ CREATE TABLE IF NOT EXISTS players (
     "timecreated" : 1332289262,
   */
 );
-CREATE INDEX IF NOT EXISTS players_cheese_idx on players(cheese) WHERE cheese IS NOT NULL AND cheese > 0;
+CREATE INDEX IF NOT EXISTS players_cheese_idx on players(cheese) WHERE cheese IS NOT NULL;
```
howardchung: undo this

howardchung: it looks like you put your changes on an old version of this file, pull and try again

store/queries.js
```diff
@@ -70,6 +70,18 @@ function cleanRowCassandra(cassandra, table, row, cb) {
...
 }
 
+function getAPIKeys(db, cb) {
+  db.raw(`
+    SELECT api_key from api_keys WHERE api_key is not null
+    `).asCallback((err, result) => {
+    if (err) {
+      return cb(err);
+    }
+    console.log(result.rows);
```
howardchung: remove logging

svc/apiadmin.js
```diff
@@ -0,0 +1,51 @@
...
+            if (cursor !== '0') {
+              storeUsageCounts(cursor);
+            }
+          });
+      },
+    );
+  });
+}
+
+setInterval(() => storeUsageCounts(0), 10 * 60 * 1000); // Every 10 minutes
```
howardchung: you may want to use invokeInterval, that causes the time to only start running once the function calls back.

with setInterval, if the job stalls, a new one will be started and it may lead to creating a lot of simultaneous stalled jobs.

svc/web.js
```diff
@@ -68,27 +85,56 @@ app.use('/ugc', (req, res) => {
...
+    rateLimit = config.API_KEY_PER_MIN_LIMIT;
+    console.log('[KEY] %s visit %s, ip %s', requestAPIKey, req.originalUrl, ip);
+  } else {
+    rateKey = `rate_limit:${ip}`;
+    countKey = `count_limit:${ip}`;
+    rateLimit = config.NO_API_KEY_PER_MIN_LIMIT;
+    console.log('[USER] %s visit %s, ip %s', req.user ? req.user.account_id : 'anonymous', req.originalUrl, ip);
+  }
+
+  redis.multi()
```
howardchung: we may want to use redis hashes for more efficient storage

sql/create_tables.sql
howardchung: was this addressed?

sql/create_tables.sql
howardchung: was this addressed?

svc/web.js
howardchung: we can store the values in redis hash/set and check for membership on each call.  that would scale better

config.js
albertcui: I was following the above line:  ENABLE_RANDOM_MMR_UPDATE: '',

svc/apiadmin.js
albertcui: I wasn't sure how to use invoke interval, especially passing in a variable (cursor). :P

config.js
howardchung: yeah but in this case you can avoid adding 2 values

albertcui: A config variable shouldn't have multiple meanings. 0 would mean there are no free API calls.

howardchung: ok, but add a comment to it

routes/spec.js
```diff
@@ -3990,6 +3990,81 @@ Please keep request rate to approximately 3/s.
...
+                FROM (  
+                  SELECT
+                    account_id,
+                    api_key,
+                    MAX(usage_count) as usage_count
+                  FROM api_key_usage
+                  WHERE
+                    timestamp >= '${startTime}'
+                    AND timestamp <= '${endTime}'
+                  GROUP BY 1, 2
```
howardchung: generally bad practice to group by numbers since it breaks the query if the select list is modified.

routes/spec.js
```diff
@@ -3990,6 +3990,81 @@ Please keep request rate to approximately 3/s.
...
+          const startTime = moment().startOf('month').format("YYYY-MM-DD");
+          const endTime = moment().endOf('month').format("YYYY-MM-DD");
+
+          async.parallel({
+            timeUsage: (cb) => {
+              cb();
+            },
+            topApi: (cb) => {
+              db.raw(`
+                SELECT
```
howardchung: isn't each account_id associated with a specific api_key?  Why not just group by account_id, api_key?

Not sure why an inner query is needed

routes/spec.js
```diff
@@ -3990,6 +3990,81 @@ Please keep request rate to approximately 3/s.
...
+                  GROUP BY 1, 2
+                ) as T1
+                GROUP BY 1
+                ORDER BY 2 DESC
+              `)
+              .asCallback((err, res) => cb(err, err ? null : res.rows));
+            },
+            topUsers: (cb) => {
+              db.raw(`
+                SELECT
```
howardchung: similarly here you could just group by api_key, ip

sql/create_tables.sql
```diff
@@ -182,6 +182,37 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+  PRIMARY KEY(account_id),
+  account_id bigint UNIQUE,
+  api_key uuid UNIQUE,
+  customer_id varchar(255)
+);
+CREATE INDEX IF NOT EXISTS api_keys_account_id_idx on api_keys(account_id);
+CREATE INDEX IF NOT EXISTS api_keys_customer_id_idx on api_keys(customer_id);
+CREATE INDEX IF NOT EXISTS api_keys_api_key_idx on api_keys(api_key);
+
+CREATE TABLE IF NOT EXISTS api_key_usage (
```
howardchung: create primary keys on these tables, account_id, api_key, timestamp?

svc/apiadmin.js
```diff
@@ -0,0 +1,98 @@
...
+      
+      if (e.startsWith('API')) {
+        db.from('api_keys').where({
+          api_key: e.replace('API:', ''),
+        }).asCallback((err, results) => {
+          if (err) {
+            cb2(err);  
+          }
+          
+          db('api_key_usage')
```
howardchung: don't you need to store timestamp as well?

moment().startOf('month')?

svc/web.js
```diff
@@ -17,11 +17,30 @@ const passport = require('passport');
...
-const redisCount = utility.redisCount;
+const { redisCount } = utility;
 
 const app = express();
 const apiKey = config.STEAM_API_KEY.split(',')[0];
 const host = config.ROOT_URL;
+
+let OD_API_KEYS = {};
+
+utility.invokeInterval((cb) => {
```
howardchung: I think we should write a background task that loads the keys into Redis from postgres (a redis set, probably), and then do a check in redis on each request if the request contains an API key.

sql/create_tables.sql
albertcui: Can't do it on user_usage since account_id could be null.

config.js
albertcui: done

config.js
albertcui: done

routes/spec.js
albertcui: It's possible for a user to delete a key and then get a new one.

routes/spec.js
albertcui: fixed

svc/web.js
albertcui: done

sql/create_tables.sql
albertcui: done

sql/create_tables.sql
albertcui: yes

config.js
```diff
@@ -56,11 +56,16 @@ const defaults = {
...
   MAXIMUM_AGE_SCENARIOS_ROWS: 2, // maximum allowed age of scenarios rows in weeks
   MATCH_CACHE_SECONDS: 60, // number of seconds to cache matches
   PLAYER_CACHE_SECONDS: 3600, // number of seconds to cache player aggregations
   SCANNER_PLAYER_PERCENT: 10, // percent of matches from scanner to insert player account IDs for (discover new player account IDs)
+  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication,
+  ENABLE_API_LIMIT: '', // if truthy API calls after exceeding API_FREE_LIMIT
+  API_FREE_LIMIT: 25000, // number of api requests per month before 429 is returned. If using an API key, calls over this are charged.
+  API_KEY_PER_MIN_LIMIT: 180, // Rate limit per minute if using an API key
+  NO_API_KEY_PER_MIN_LIMIT: 60, // RAte limit per minute if not using an API key
+  ADMIN_ACCOUNT_IDS: []
```
howardchung: add a comment

config.js
```diff
@@ -56,11 +56,16 @@ const defaults = {
   ENABLE_PLAYER_CACHE: 1, // enable/disable player aggregation caching
   ENABLE_RANDOM_MMR_UPDATE: '', // set to request MMR updates after ranked matches
   WEBSOCKET_PORT: 80, // port for live match subscription websocket service
-  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication
   MAXIMUM_AGE_SCENARIOS_ROWS: 2, // maximum allowed age of scenarios rows in weeks
   MATCH_CACHE_SECONDS: 60, // number of seconds to cache matches
   PLAYER_CACHE_SECONDS: 3600, // number of seconds to cache player aggregations
   SCANNER_PLAYER_PERCENT: 10, // percent of matches from scanner to insert player account IDs for (discover new player account IDs)
+  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication,
+  ENABLE_API_LIMIT: '', // if truthy API calls after exceeding API_FREE_LIMIT
```
howardchung: "block API calls"?

sql/create_tables.sql
```diff
@@ -182,6 +182,33 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+  account_id bigint REFERENCES api_keys(account_id),
+  customer_id varchar(255),
+  api_key uuid,
+  usage_count bigint,
+  ip cidr,
+  timestamp timestamp default current_timestamp
+);
+CREATE INDEX IF NOT EXISTS api_keys_usage_account_id_idx on api_keys(account_id);
+
+CREATE TABLE IF NOT EXISTS user_usage (
```
howardchung: ip, timestamp as primary key?

routes/spec.js
```diff
@@ -3990,6 +3990,177 @@ Please keep request rate to approximately 3/s.
...
+        },
+        route: () => '/admin/apiMetrics',
+        func: (req, res, cb) => {
+          const startTime = moment().startOf('month').format('YYYY-MM-DD');
+          const endTime = moment().endOf('month').format('YYYY-MM-DD');
+
+          async.parallel({
+            timeAPIUsage: (cb) => {
+              db.raw(`
+                SELECT
```
howardchung: If the API key is mutable, you could just group by account_id (assuming a single account ID cannot have more than one API key at a time)

svc/apiadmin.js
```diff
@@ -0,0 +1,96 @@
...
+            cb2();
+          }
+        });
+      } else if (e.startsWith('USER')) {
+        const split = e.split(':');
+
+        db('user_usage')
+          .insert({
+            ip: split[1],
+            account_id: split[2] ? split[2] : null,
```
howardchung: split[2] || null?

svc/apiadmin.js
```diff
@@ -0,0 +1,96 @@
...
+          console.log(res);
+          cb();
+        });
+    } else {
+      cb();
+    }
+  });
+}, 5 * 60 * 1000); // Update every 5 min
+
+invokeInterval(cb => storeUsageCounts(0, cb), 10 * 60 * 1000); // Every 10 minutes
```
howardchung: won't this create a new row every 10 minutes?  You only want one row per billing period (month)?

svc/web.js
```diff
@@ -68,27 +69,66 @@ app.use('/ugc', (req, res) => {
...
+      } else {
+        res.locals.isAPIRequest = resp === 1;
+        cb();
+      }
+    });
+  } else {
+    cb();
+  }
+});
 app.use((req, res, cb) => {
```
howardchung: we should consider adding a test for the limiting logic

sql/create_tables.sql
albertcui: I don't think we can do this because multiple accounts might be on the same IP.

svc/apiadmin.js
albertcui: done

svc/apiadmin.js
albertcui: updated

sql/create_tables.sql
albertcui: Added a unique index. Also mapping null to 0 to avoid duplicate rows since null is not equal to itself.

svc/web.js
albertcui: I tried adding tests but I can't get them to run locally so no idea if it works :P

albertcui: Test is now passing.

routes/spec.js
```diff
@@ -3990,6 +3990,163 @@ Please keep request rate to approximately 3/s.
...
+                  T1.api_key as api_key,
+                  MAX(api_key_usage.usage_count) as usage_count
+                FROM api_key_usage
+                JOIN ( 
+                  SELECT
+                    api_key,
+                    MAX(usage_count) as usage_count
+                  FROM api_key_usage
+                  WHERE
+                    timestamp >= '${startTime}'
```
howardchung: parameterize the query instead of string interpolation.  I know this is not user input, but it's still safer to avoid accidental SQL injection

routes/spec.js
```diff
@@ -3990,6 +3990,163 @@ Please keep request rate to approximately 3/s.
         },
       },
     },
+    '/admin/apiMetrics': {
```
howardchung: what is the plan to consume these metrics?  are you adding to the status page/UI?

sql/create_tables.sql
```diff
@@ -182,6 +182,34 @@ CREATE TABLE IF NOT EXISTS subscriptions (
...
+);
+CREATE INDEX IF NOT EXISTS api_keys_account_id_idx on api_keys(account_id);
+
+CREATE TABLE IF NOT EXISTS api_key_usage (
+  PRIMARY KEY(account_id, api_key, timestamp),
+  account_id bigint REFERENCES api_keys(account_id),
+  customer_id varchar(255),
+  api_key uuid,
+  usage_count bigint,
+  ip cidr,
```
howardchung: i have noticed that the IP address is not always well-formed (ipv6 or sometimes weird values).  You may want to check the logs to see what kinds of IPs we are getting before deploying this.

Maybe safer to use a text type.

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+          api_key: split[2],
+        }).asCallback((err, results) => {
+          if (err) {
+            cb2(err);
+          }
+          if (results.length > 0) {
+            db.raw(`
+              INSERT INTO api_key_usage
+              (account_id, api_key, customer_id, timestamp, ip, usage_count) VALUES
+              (${results[0].account_id},
```
howardchung: parameterize the query

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+      }
+    });
+  });
+}
+
+
+utility.invokeInterval((cb) => {
+  queries.getAPIKeys(db, (err, rows) => {
+    if (err) {
+      cb();
```
howardchung: return cb()?

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+           '${split[1]}',
+           ${values[i + 1]})
+          ON CONFLICT (account_id, ip, timestamp) DO UPDATE SET usage_count = ${values[i + 1]}
+        `)
+          .asCallback(cb2);
+      }
+    }, (err) => {
+      if (err) {
+        console.error('[ERROR] ', err);
+        process.exit(1);
```
howardchung: why kill the process instead of just cb(err)?

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+utility.invokeInterval((cb) => {
+  queries.getAPIKeys(db, (err, rows) => {
+    if (err) {
+      cb();
+    }
+
+    if (rows.length > 0) {
+      const keys = rows.map(e => e.api_key);
+
+      redis.multi()
```
howardchung: is multi guaranteed to be atomic?

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+        .del('api_keys')
+        .sadd('api_keys', keys)
+        .exec((err, res) => {
+          if (err) {
+            console.error('[ERROR] ', err);
+            process.exit(1);
```
howardchung: same comment about kill vs cb(err)

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+
+      if (cursor !== '0') {
+        storeUsageCounts(cursor, cb);
+      }
+    });
+  });
+}
+
+
+utility.invokeInterval((cb) => {
```
howardchung: in general I try to avoid having a service do two timed loops.  If you feel that these operations are very closely related it's OK, but debugging may be more difficult.

routes/spec.js
```diff
@@ -3990,6 +3990,163 @@ Please keep request rate to approximately 3/s.
...
+            },
+            topAPI: (cb) => {
+              db.raw(`
+                SELECT
+                  account_id,
+                  SUM(usage_count) as usage_count,
+                  ARRAY_AGG(api_key) as api_keys,
+                  ARRAY_AGG(DISTINCT ip) as ips
+                FROM (  
+                  SELECT
```
howardchung: why do you need the inner query here?  Aren't you just getting the top users for the month (so just pick up the rows with the largest timestamp?)

svc/web.js
```diff
@@ -68,27 +69,66 @@ app.use('/ugc', (req, res) => {
 app.use(session(sessOptions));
 app.use(passport.initialize());
 app.use(passport.session());
-// Rate limiter middleware
+
+// Rate limiter and API key middleware
+app.use((req, res, cb) => {
+  if (req.query.OPENDOTA_API_KEY) {
```
howardchung: OPENDOTA_API_KEY should not be the query param, something like `key`

test/test.js
```diff
@@ -230,14 +230,65 @@ describe('api', () => {
...
+      .exec((err) => {
+        if (err) {
+          return done(err);
+        }
+
+        return done();
+      });
+  });
+
+  it('non api key calls should be able to make 20. 21st should fail.', (done) => {
```
howardchung: start the string with should

routes/spec.js
albertcui: If redis dies the largest timestamp is not guaranteed to have the largest usage_count.

svc/web.js
albertcui: Changed to API_KEY :P

svc/apiadmin.js
albertcui: Changed.

svc/apiadmin.js
albertcui: Changed.

test/test.js
albertcui: Done

svc/apiadmin.js
albertcui: You get annoying lint errors "Expected to return a value at the end of arrow function  consistent-return"

I wrapped the following code in an else block.

routes/spec.js
howardchung: If Redis dies, won't the next run (10 minutes later) write the latest data?

svc/apiadmin.js
howardchung: consistent return just means you should return something at the end.  You can return null

svc/web.js
howardchung: the rest of our query params are lower case, I think we should stay consistent

svc/apiadmin.js
```diff
@@ -0,0 +1,109 @@
...
+
+    const cursor = results[0];
+    const values = results[1];
+
+    const apiTimestamp = moment().startOf('day');
+    const userTimestamp = moment().startOf('month');
+
+    async.eachOfLimit(values, 5, (e, i, cb2) => {
+      if (i % 2) {
+        cb2();
```
howardchung: return cb2()?

routes/spec.js
howardchung: update on this question?

sql/create_tables.sql
howardchung: did you validate that the ips are well-formed?

svc/apiadmin.js
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      const cursor = results[0];
+      const values = results[1];
+
+      const apiTimestamp = moment().startOf('day');
+      const userTimestamp = moment().startOf('month');
+
+      async.eachOfLimit(values, 5, (e, i, cb2) => {
+        if (i % 2) {
+          cb2();
```
howardchung: return?

svc/apiadmin.js
```diff
@@ -0,0 +1,105 @@
...
+            api_key: split[2],
+          }).asCallback((err, results) => {
+            if (err) {
+              cb2(err);
+            }
+            if (results.length > 0) {
+              db.raw(`
+                INSERT INTO api_key_usage
+                (account_id, api_key, customer_id, timestamp, ip, usage_count) VALUES
+                (${results[0].account_id},
```
howardchung: parameterize query

howardchung: if you don't like to use ? you can do it with named parameters as well

svc/apiadmin.js
```diff
@@ -0,0 +1,105 @@
...
+            }
+          });
+        } else if (e.startsWith('USER')) {
+          const split = e.split(':');
+
+          // null account_id mapped to 0 to avoid duplicate rows
+          db.raw(`
+            INSERT INTO user_usage
+            (account_id, timestamp, ip, usage_count) VALUES
+            (${split[2] || 0},
```
howardchung: parameterize

howardchung:
Before deploying, the DB schema needs to be updated.

routes/spec.js
albertcui: There is the possibility that Redis dies and we lose the key.

howardchung: redis persists to disk every 1 or 5 minutes right now depending on how many keys changed.  Are you worried about losing 5 minutes of data?

config.js
```diff
@@ -56,11 +56,16 @@ const defaults = {
...
   MAXIMUM_AGE_SCENARIOS_ROWS: 2, // maximum allowed age of scenarios rows in weeks
   MATCH_CACHE_SECONDS: 60, // number of seconds to cache matches
   PLAYER_CACHE_SECONDS: 3600, // number of seconds to cache player aggregations
   SCANNER_PLAYER_PERCENT: 10, // percent of matches from scanner to insert player account IDs for (discover new player account IDs)
+  ENABLE_RETRIEVER_ADVANCED_AUTH: '', // set to enable retriever two-factor and SteamGuard authentication,
+  ENABLE_API_LIMIT: '', // if truthy, API calls after exceeding API_FREE_LIMIT are blocked
+  API_FREE_LIMIT: 25000, // number of api requests per month before 429 is returned. If using an API key, calls over this are charged.
+  API_KEY_PER_MIN_LIMIT: 180, // Rate limit per minute if using an API key
+  NO_API_KEY_PER_MIN_LIMIT: 60, // Rate limit per minute if not using an API key
+  ADMIN_ACCOUNT_IDS: [], // Whitelisted account IDs to access /admin* routes
```
howardchung: actually you have to use a string here since that's what we can pass through dotenv.  Make this a comma delimited string and then split it into an array at runtime

howardchung: and indicate in the comment what the delimiter is

sql/create_tables.sql
albertcui: I checked logs and did not see anything weird. cidr supports IPV6 anyway.

howardchung: ok, but ensure that the API request does not fail if we fail this telemetry insert

howardchung: looks like this is inserted into in the apiadmin service so just monitor that for crashes/errors I guess

albertcui: Changed to text.

svc/apiadmin.js
albertcui: fixed

svc/apiadmin.js
albertcui: fixed

svc/apiadmin.js
albertcui: fixed

config.js
albertcui: I tried an array and it works?

.env contains:

ADMIN_ACCOUNT_IDS =[102344608, 123]

howardchung: it might depending on how dotenv parses it.  we're using comma delimited values for other things like steam API keys though so might be better to be consistent

albertcui: This was your idea :P

howardchung: your original version just used includes on the string directly?  The difference is that you should split the string into an array first

routes/spec.js
albertcui: IDK how to write this query then

                  SELECT
                    account_id,
                    ARRAY_AGG(api),
                    ARRAY_AGG(ip),
                    SUM(usage_count) as usage_count,
                    MAX(timestamp)
                  FROM api_key_usage
                  WHERE
                    timestamp >= ?
                    AND timestamp <= ?
                  GROUP BY account_id, api_key, ip

howardchung:
```
select
account_id,
api,
ip,
usage_count
from api_key_usage
where timestamp = (select max(timestamp) from api_key_usage)
```

howardchung: you may want an index on timestamp to find the max quickly

routes/spec.js
```diff
@@ -4121,6 +4121,143 @@ Please keep request rate to approximately 3/s.
...
+                    api_key,
+                    usage_count
+                  FROM api_key_usage
+                  WHERE
+                    timestamp = (SELECT MAX(timestamp) FROM api_key_usage)
+                  ORDER BY usage_count DESC
+                  LIMIT 50
+                ) as T1
+                on api_key_usage.api_key = T1.api_key
+                GROUP BY day, T1.api_key
```
howardchung: I don't think you need the outer query anymore

svc/apiadmin.js
howardchung: not in the latest commit

routes/spec.js
albertcui: This one needs it because it's getting the time-series data of the top 50 API keys.

svc/apiadmin.js
albertcui: This is fixed with an else if instead of a return.

routes/spec.js
albertcui: Yes it will go into UI eventually, but that should be a follow up to this PR.

routes/spec.js
howardchung: In that case the inner query won't work since it's only getting the most recent data.  I wasn't sure what you were trying to do

svc/apiadmin.js
howardchung: I don't see an else?

routes/spec.js
albertcui: The inner query identifies the top API keys. The outer query get the time series data, limiting to the top 50 API keys via the join. I think this works?

howardchung: Ok that should be fine

albertcui:
Deployed schema change.

svc/apiadmin.js
albertcui: I think it's because you commented on a specific commit: https://github.com/odota/core/pull/1622/files#diff-a28999e9c7ae146209dadff59251f966

PR #1627, opened 2018-04-05
Scenarios metadata endpoint. removed pings scenarios. row age.

Argn0:
I removed the pings scenario. The data was not very interesting.


routes/spec.js
```diff
@@ -4074,6 +4074,28 @@ Please keep request rate to approximately 3/s.
         },
       },
     },
+    '/scenarios/metaData': {
```
howardchung: i think our convention is to use all lower case (also metadata is one word)

howardchung: actually we do use camelCase but metadata is still one word

routes/spec.js
```diff
@@ -4074,6 +4074,28 @@ Please keep request rate to approximately 3/s.
...
+          200: {
+            description: 'Success',
+            schema: {
+              type: 'object',
+            },
+          },
+        },
+        route: () => '/scenarios/metaData',
+        func: (req, res, cb) => {
+          queries.getScenariosMetaData(data => res.json(data), cb);
```
howardchung: I think you can just use the existing metadata endpoint rather than implementing a new one.

sql/create_tables.sql
```diff
@@ -456,13 +456,11 @@ CREATE TABLE IF NOT EXISTS scenarios (
   hero_id smallint,
   item text,
   time integer,
-  pings integer,
   lane_role smallint,
   games bigint DEFAULT 1,
   wins bigint,
   epoch_week integer,
   UNIQUE (hero_id, item, time, epoch_week),
-  UNIQUE (pings, time, epoch_week),
```
howardchung: I think this constraint needs to be dropped first in migrations or postgres will not allow the column drop

store/queries.js
```diff
@@ -1078,6 +1078,31 @@ function getTeamScenarios(req, cb) {
...
+      }));
+    },
+    user(cb) {
+      cb(null, req.user);
+    },
+  }, (err, result) => {
+    if (err) {
+      return cb(err);
+    }
+    return res.json(result);
```
howardchung: I would not advise doing res.json here, these functions should return objects/arrays so they're reusable.  Keep res.json in the spec.js file.

util/scenariosUtil.js
```diff
@@ -148,6 +132,10 @@ const scenarioChecks = {
 // list of match object properties that are required for scenario checks.
 const matchProperties = ['players', 'objectives', 'duration', 'chat', 'radiant_win'];
 
+const metadata = {
+  itemCost, timings, gameDurationBucket, teamScenariosQueryParams,
```
howardchung: one per line

util/scenariosUtil.js
```diff
@@ -8,7 +8,6 @@ const playerWon = utility.playerWon;
 const itemCost = 2000;
 const dotaItems = Object.keys(constants.items).map(k => [constants.items[k], k]).filter(x => x[0].cost >= itemCost).map(x => x[1]);
 const timings = [7.5, 10, 12, 15, 20, 25, 30].map(x => x * 60);
-const pingBucket = [10, 25, 50, 100, 150, 200, 500, 1000];
 const gameDurationBucket = [15, 30, 45, 60, 90].map(x => x * 60);
 
 const negativeWords = ['ff', 'report', 'gg', 'end', 'noob'];
```
howardchung: I actually think we should swap to using internal strings for teamScenariosQueryParams, e.g. `pos_chat_1min`.  Then pass these down to web, and put the translated strings there so the translators can localize it.

Argn0:
-pass callback to queries.getMetaData
-use internal strings for teamScenariosQueryParams
-update team_scenarios table with internal strings

Issue #1630, opened 2018-04-10
scenarios error


howardchung:
@Argn0, found this in the logs:
```
18|scenari | TypeError: Cannot read property 'forEach' of null
18|scenari |     at match.players.forEach (/usr/src/util/scenariosUtil.js:38:29)
18|scenari |     at Array.forEach (<anonymous>)
18|scenari |     at itemTimings (/usr/src/util/scenariosUtil.js:37:21)
18|scenari |     at su.scenarioChecks.(anonymous function).forEach (/usr/src/svc/scenarios.js:21:22)
18|scenari |     at Array.forEach (<anonymous>)
18|scenari |     at Object.keys.forEach (/usr/src/svc/scenarios.js:20:32)
18|scenari |     at Array.forEach (<anonymous>)
18|scenari |     at buildMatch (/usr/src/svc/scenarios.js:19:36)
18|scenari |     at Command.redis.setex [as callback] (/usr/src/store/buildMatch.js:178:18)
18|scenari |     at normal_reply (/usr/src/node_modules/redis/index.js:726:21)
18|scenari |     at RedisClient.return_reply (/usr/src/node_modules/redis/index.js:824:9)
18|scenari |     at JavascriptRedisParser.returnReply (/usr/src/node_modules/redis/index.js:192:18)
18|scenari |     at JavascriptRedisParser.execute (/usr/src/node_modules/redis-parser/lib/parser.js:574:12)
18|scenari |     at Socket.<anonymous> (/usr/src/node_modules/redis/index.js:274:27)
18|scenari |     at emitOne (events.js:116:13)
18|scenari |     at Socket.emit (events.js:211:7)
```

PR #1632, opened 2018-04-10
Fix scenarios error and some refactoring

Argn0:
fixes https://github.com/odota/core/issues/1630

### DIFF
--- a/util/scenariosUtil.js
+++ b/util/scenariosUtil.js
@@ -4,8 +4,8 @@ const utility = require('./utility');
 const playerWon = utility.playerWon;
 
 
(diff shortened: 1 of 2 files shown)


util/scenariosUtil.js
```diff
@@ -4,8 +4,8 @@ const utility = require('./utility');
 const playerWon = utility.playerWon;
 
 
-// all items that cost at least 2000
-const itemCost = 2000;
+// all items that cost at least 1400
```
howardchung: optional: I feel like putting the constant value in the comment is kind of redundant :)

PR #1644, opened 2018-04-25
whitelist routes from api

### LINKED ISSUE #1643: exempt informational endpoints from monthly limit
e.g. keys, metadata.

Otherwise users who exceed the monthly rate limit are unable to do anything, including sign up for an API key.

@albertcui

albertcui:
fixes #1643

### DIFF
--- a/svc/web.js
+++ b/svc/web.js
@@ -166,7 +166,8 @@ app.use((req, res, cb) => {
       console.log('[SLOWLOG] %s, %s', req.originalUrl, elapsed);
     }
 
-    if (res.statusCode !== 500 && !whitelistedPaths.includes(req.path)) {
+    // When called from a middleware, the mount point is not included in req.path. See Express docs.
+    if (res.statusCode !== 500 && !whitelistedPaths.includes(req.baseUrl + (req.path === '/' ? '' : req.path))) {
       redis.multi()
         .hincrby('usage_count', res.locals.usageIdentifier, 1)
         .expireat('usage_count', utility.getEndOfMonth())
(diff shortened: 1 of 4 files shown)


svc/web.js
```diff
@@ -29,6 +29,16 @@ const sessOptions = {
   maxAge: 52 * 7 * 24 * 60 * 60 * 1000,
   secret: config.SESSION_SECRET,
 };
+
+const whitelistedPaths = [
+  '/api', // Docs
+  '/api/metadata', // Login status
+  '/login',
```
howardchung: think passport routes are before the rate limiter anyway so it shouldn't matter

albertcui: they won't get 429'd but they'll get API incremented.

svc/web.js
```diff
@@ -29,6 +29,16 @@ const sessOptions = {
   maxAge: 52 * 7 * 24 * 60 * 60 * 1000,
   secret: config.SESSION_SECRET,
 };
+
+const whitelistedPaths = [
```
howardchung: might want to refactor this into its own file and then reference it in both web and test

svc/web.js
```diff
@@ -115,7 +125,7 @@ app.use((req, res, cb) => {
     rateLimit = config.API_KEY_PER_MIN_LIMIT;
     console.log('[KEY] %s visit %s, ip %s', requestAPIKey, req.originalUrl, ip);
   } else {
-    res.locals.usageIdentifier = `USER:${ip}:${req.user ? req.user.account_id : ''}`;
+    res.locals.usageIdentifier = `USER:${ip}`;
```
howardchung: will this change effectively reset everyone's rate limits?

albertcui: yes, but there's only a couple of days usage and this month is almost over anyway.

PR #1648, opened 2018-05-09
add rate limit information headers to api calls

bippum:


howardchung:
@albertcui do you want to review this?

howardchung:
I believe these failures are due to Travis CI not setting the test stripe credential.  @albertcui can confirm.  possibly we should just hardcode in the test stripe key if it's not sensitive.

svc/web.js
```diff
@@ -139,6 +139,13 @@ app.use((req, res, cb) => {
...
 
   multi.exec((err, resp) => {
+    res.set({
+      'X-Rate-Limit': rateLimit,
+      'X-Rate-Limit-Remaining': rateLimit - resp[0],
+    });
+    if (!res.locals.isAPIRequest) {
+      res.set('X-Free-Requests-Remaining', config.API_FREE_LIMIT - Number(resp[2]));
+    }
     if (err) {
```
albertcui: We should set the header after the err check. Otherwise, if there was an error, the resp information won't be accurate.

albertcui:
Thanks! Merging.

PR #1653, opened 2018-05-28
fix incorrect api limit exceeded error

bippum:
On some paths i am getting `monthly api limit reached` even though i am using a key and below 50,000 requests. Changing these two conditionals should make the check for an API key succeed before checking if a path is whitelisted or not.

### DIFF
--- a/svc/web.js
+++ b/svc/web.js
@@ -159,7 +159,7 @@ app.use((req, res, cb) => {
         error: 'rate limit exceeded',
       });
     }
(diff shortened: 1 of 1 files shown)


howardchung:
which path are you getting this on?

bippum:
nevermind was user error. https://github.com/bippum/mika/commit/2158104cf5d96f26c807ed4a281f95e65916f7f8

PR #1656, opened 2018-06-04
fix(package): update stripe to version 6.1.0

albertcui:
Closes #1650


howardchung:
don't you need to update package-lock.json

albertcui:
all I did was use greenkeeper

howardchung:
greenkeeper doesn't update the lockfile, it's more useful just as a notification when a new package is released

PR #1657, opened 2018-06-04
[WIP] Notifications for new matches

albertcui:


svc/parser.js
```diff
@@ -61,9 +61,34 @@ function parseProcessor(job, cb) {
       console.error(err.stack || err);
     } else {
       console.log('completed parse of match %s', match.match_id);
+      sendNotifications(match);
     }
     return cb(err, match.match_id);
   });
 }
 
+function sendNotifications(match) {
```
howardchung: you probably have to check the origin of the parse request and do this for scanner only (similar to what was done for scenarios), otherwise, people requesting parses might cause multiple notifications?

routes/notificationManagement.js
```diff
@@ -0,0 +1,40 @@
...
+  if (!req.user) {
+    return res.status(403).json({
+      error: 'Authentication required',
+    });
+  }
+
+  return next();
+});
+
+notifications.route('/')
```
howardchung: what is this route used for

svc/parser.js
```diff
@@ -61,9 +61,34 @@ function parseProcessor(job, cb) {
       console.error(err.stack || err);
     } else {
       console.log('completed parse of match %s', match.match_id);
+      sendNotifications(match);
```
howardchung: I think we should make this only enabled if a config option is set.  probably there are people using this that don't need this feature

routes/notificationManagement.js
albertcui: It's used to store the notification token on the server so that we know who to send notifications to.

svc/parser.js
albertcui: code pointer?

howardchung: https://github.com/odota/core/blob/master/svc/parser.js#L39

you could add something like doNotifications as a property on the match, or just check directly

PR #1661, opened 2018-06-07
[WIP] Elasticsearch for Search

albertcui:


store/elasticsearch.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Interface to PostgreSQL client
```
howardchung: incorrect comment :)

howardchung:
do we also need a service that transfers data from players table to elasticsearch?

albertcui:
Not if we roll it out in parallel to current search.

PR #1664, opened 2018-06-19
Benchmark data from parsed matches 

Argn0:
https://github.com/odota/web/issues/524


svc/counts.js
```diff
@@ -225,7 +200,7 @@ function processCounts(match, cb) {
     },
     updateBenchmarks(cb) {
       if (isSignificant(match)) {
-        return updateBenchmarks(match, cb);
+        return benchmarksUtil.updateBenchmarks(match, cb);
```
howardchung: perhaps we should take this out and just let the separate service do the benchmarking?

util/benchmarksUtil.js
```diff
@@ -0,0 +1,70 @@
...
+  hero_healing_per_min(m, p) {
+    return (p.hero_healing / m.duration) * 60;
+  },
+  tower_damage(m, p) {
+    return p.tower_damage;
+  },
+};
+
+const parsedBenchmarks = {
+  stuns(m, p) {
```
howardchung: probably should just combine these and benchmark parsed matches only.  doing all matches for some and parsed matches for others seems too complicated

Argn0:
I made the changes you suggested.

config.js
```diff
@@ -52,6 +52,7 @@ const defaults = {
   SCANNER_PERCENT: 100, // percent of matches to insert from scanner
   PUBLIC_SAMPLE_PERCENT: 10, // percent of public matches to sample in DB
   SCENARIOS_SAMPLE_PERCENT: 50, // percent of parsed matches to sample for scenarios
+  BENCHMARKS_SAMPLE_PERCENT: 100, // percent of parsed matches to sample for benchmarks
```
howardchung: since the number of parsed matches is lower, we should probably increase BENCHMARK_RETENTION_MINUTES to an hour or something

Argn0:
Let me do some tests. I might have broken something.

Argn0:
OK, all is good now.

howardchung:
Could you also add a metric for the queue length to buildStatus (like you did for scenariosQueue?)

PR #1665, opened 2018-06-19
Fix missing path parameters in API spec

13k:
Add missing path parameters in API spec


howardchung:
I believe there's a known issue with the API metering tests failing due to a required env var in Travis, @albertcui is aware of it.  

PR looks good

PR #1670, opened 2018-06-23
Change stuns per minute

Argn0:
It just occurred to me, that the value probably should be per minute.

### DIFF
--- a/util/benchmarksUtil.js
+++ b/util/benchmarksUtil.js
@@ -20,7 +20,7 @@ const benchmarks = {
   tower_damage(m, p) {
     return p.tower_damage;
   },
-  stuns(m, p) {
+  stuns_per_min(m, p) {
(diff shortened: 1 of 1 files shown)


util/benchmarksUtil.js
```diff
@@ -21,7 +21,7 @@ const benchmarks = {
     return p.tower_damage;
   },
   stuns(m, p) {
```
howardchung: do we want to change this too?

Argn0: I dont think so. Because there is only a limited amount of tower damage you can do per game.

howardchung: oh, i was referring to the key name for stuns.  github doesn't highlight the line very well

Argn0: Oh yea, my bad actually.

PR #1682, opened 2018-07-15
add rollout flags for es

albertcui:


routes/spec.js
```diff
@@ -60,6 +60,19 @@ function sendDataWithCache(req, res, data, key) {
   return res.json(data);
 }
 
+
+function hash(s) {
```
howardchung: can't you use the prebuilt hash functions in node instead of implementing this?

util/utility.js
```diff
@@ -805,6 +806,10 @@ function cleanItemSchema(input) {
   return input;
 }
 
+function checkIfInExperiment(ip, mod) {
+  return crypto.createHash('md5').update(ip).digest('hex') % 100 < mod;
```
howardchung: isn't the hex digest a string?  does mod work on it?

PR #1731, opened 2018-09-18
Add Contributor state to API 

gu3st:
Created contributors file for OpenDota contributors to flag their profile if desired.  Added it to the API response for /players/:id.


CONTRIBUTORS.js
```diff
@@ -0,0 +1,9 @@
+/*
```
howardchung: might be easier to just make this a JSON file

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
```
howardchung: don't think package-lock needs to be changed

routes/spec.js
```diff
@@ -1017,6 +1018,11 @@ Please keep request rate to approximately 1/s.
                       description: 'loccountrycode',
                       type: 'string',
                     },
+                    is_contributor: {
+                      description: 'Has the user contributed to the development of OpenDota',
```
howardchung: I think better to start with "Boolean indicating if the user has contributed..."

routes/spec.js
```diff
@@ -1028,7 +1034,12 @@ Please keep request rate to approximately 1/s.
           const accountId = Number(req.params.account_id);
           async.parallel({
             profile(cb) {
-              queries.getPlayer(db, accountId, cb);
+              queries.getPlayer(db, accountId, (err, playerData) => {
+                if (!err && playerData !== undefined) {
```
howardchung: think you can just check if (playerData)

CONTRIBUTORS.js
howardchung: Also add:
88367253 (me)
102344608 (@albertcui)

gu3st: The only reason I made it a .js was so I could include a comment to explain the file. If JSON supported comments I would have done JSON.

howardchung: ok, that's fine :)

PR #1766, opened 2018-10-17
Update CONTRIBUTORS.js

diw1:


diw1:
was trying to type when the dota2 popup and forgot, but +25 👍

PR #1768, opened 2018-10-19
Build match refactoring

ichbinsalman:
Removed all callback functions from buildmatch.js and replaced them with asnyc await functions

### DIFF
--- a/store/buildMatch.js
+++ b/store/buildMatch.js
@@ -2,7 +2,6 @@
  * Functions to build/cache match object
  * */
 const config = require('../config');
-const async = require('async');
 const queries = require('./queries');
 const compute = require('../util/compute');
 const utility = require('../util/utility');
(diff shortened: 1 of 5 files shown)


store/buildMatch.js
```diff
@@ -2,189 +2,154 @@
...
+    radiant_team: radiantTeam,
+    dire_team: direTeam,
   });
 }
 
-function getMatch(matchId, cb) {
+async function getMatch(matchId) {
   if (!matchId || Number.isNaN(Number(matchId)) || Number(matchId) <= 0) {
-    return cb();
+    return Promise.reject();
```
howardchung: does this return 404?

test/test.js
```diff
@@ -236,22 +236,18 @@ describe('replay parse', function testReplayParse() {
...
-            assert(match.players);
-            assert(match.players[0]);
-            assert(match.players[0].killed.npc_dota_creep_badguys_melee === 46);
-            assert(match.players[0].lh_t && match.players[0].lh_t.length > 0);
-            assert(match.teamfights && match.teamfights.length > 0);
-            assert(match.draft_timings);
-            assert(match.radiant_gold_adv && match.radiant_gold_adv.length > 0);
-            return done();
-          });
+          const match = await buildMatch(tests[key].match_id);
```
howardchung: what if this errors?

PR #1793, opened 2018-11-12
feedv2 implementation

### LINKED ISSUE #1620: implement feedv2 API
@bippum is interested in working on this

* Requires Redis 5 (streams)
* Uses the Redis stream sequence number to allow users to pick up where they disconnected (no more missing matches)
* Evaluate each blob we plan to send to the client -- does it contain a given account_id?  is it a particular game mode?  is it part of a particular league?
* Each time we get a match from steam API or parse a match, push the blob into a redis stream
* on each insert, set a max size for the stream (10,000 blobs?), so old data is automatically cleaned up
* Use streaming HTTP instead of websocket for improved reliability and integration with the documentation
* Example: `https://api.opendota.com/feed?account_id=1`

howardchung:
fixes #1620 

@bippum 

This is a sample implementation of the feed using Redis streams.


package-lock.json
```diff
@@ -5336,282 +5418,63 @@
         "buffer-crc32": "^0.2",
         "bytebuffer": "^5.0",
         "steam-crypto": "^0.0",
-        "steam-resources": "git+https://github.com/seishun/node-steam-resources.git#v1.0.0"
+        "steam-resources": "git+https://github.com/seishun/node-steam-resources.git#1403939a12ba467a4b82c8ccbc279bbcdad43588"
```
howardchung: need to fix this as it seems like latest versions of npm will overwrite this. changing to this causes replay salt fetching to fail.

howardchung:
also need to disable test of this endpoint as it doesn't return a response (long-lived request)

PR #1819, opened 2018-12-03
Add connection_log to players (#1783)

### LINKED ISSUE #1783: add connection_log to match objects
Requested feature, log player connect/disconnect events (these are in the replay)

fcostantini:
fixes #1783 

Hi, I got something done for this feature (#1783). My main question would be how to distinguish a disconnection from a reconnection in the log. I just added an attribute called _event_ ~~(0 for disc, 1 for rec)~~ to the object, but of course I'm open to suggestions on how to do it better.

I still need a match where lots of disc/recs occur but here's how a single reconnection looks in the object:
image.

I'll try to get a better sample later.

~~Also the disconnections after the game is done are not logged, which I assume is intended behavior.~~

Please let me know what you think.

Cheers!

EDIT: ok, I fixed some stuff because I got the messages wrong previously. _event_ now is 0 for a "normal" disconnect, like at the end of the game, 1 for a disconnect where the player is expected to reconnect (for example your game closes in the middle of the game, or your internet goes down) and 2 for a reconnect. Please let me know if 0 and 1 should be treated as the same thing or not. Here's how it looks now (match number 4253391864):
image


processors/processExpand.js
```diff
@@ -406,6 +406,30 @@ function processExpand(entries, meta) {
       // starting by the Dire side to the Dire Side, Bottom to Top, Melee to Ranged
       // so Bottom Melee Dire Rax = 1 and Top Ranged Radiant Rax = 2048.
     },
+    CHAT_MESSAGE_RECONNECT(e) {
+      expand({
+        time: e.time,
+        type: 'connection_log',
+        slot: e.player1,
+        event: 2,
```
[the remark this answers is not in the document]
howardchung: I don't think the end of game disconnects are of interest, so you can probably ignore the third type.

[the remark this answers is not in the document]
fcostantini: Will do!

sql/create_tables.cql
```diff
@@ -78,6 +78,7 @@ CREATE TABLE IF NOT EXISTS player_matches (
   kills_log text,
   buyback_log text,
   runes_log text,
+  connection_log text,
```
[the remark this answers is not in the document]
fcostantini: Done, not sure if I got it right so let me know otherwise :-)

fcostantini:
Here's how it looks after your suggestions: 

image

Let me know what you think :)

processors/processLogParse.js
```diff
@@ -36,6 +36,9 @@ function processReduce(entries, meta) {
       || e.type === 'CHAT_MESSAGE_UNPAUSED'
       || e.type === 'CHAT_MESSAGE_RUNE_PICKUP'
       || e.type === 'CHAT_MESSAGE_FIRSTBLOOD'
+      || e.type === 'CHAT_MESSAGE_DISCONNECT_WAIT_FOR_RECONNECT'
```
howardchung: I don't think CHAT_MESSAGE_DISCONNECT is needed here

routes/spec.js
```diff
@@ -376,6 +376,24 @@ The OpenDota API provides Dota 2 related data including advanced match data extr
...
                         type: 'integer',
                       },
+                      connection_log: {
+                        description: 'Array containing information about the player\'s disconnections and reconnections',
+                        type: 'array',
+                        items: {
+                          type: 'object',
+                          properties: {
+                            time: {
+                              description: 'Time in seconds the event ocurred',
```
howardchung: I'd say "Game time in seconds"

processors/processLogParse.js
fcostantini: Yep, forgot to remove it.

routes/spec.js
fcostantini: Done

PR #1837, opened 2019-01-21
 Added `name` and `is_contributor` fields to peers

builder-247:
To allow displaying contributor and confirmed badges on web

### DIFF
--- a/store/queries.js
+++ b/store/queries.js
@@ -376,20 +376,21 @@ function getPeers(db, input, player, cb) {
   // limit to 200 max players
   teammatesArr = teammatesArr.slice(0, 200);
   return async.each(teammatesArr, (t, cb) => {
-    db.first().from('players').where({
+    db.first().from('players').leftJoin('notable_players', 'players.account_id', 'notable_players.account_id').where({
       account_id: t.account_id,
-    }).asCallback((err, row) => {
-      if (err || !row) {
+    })
+      .asCallback((err, row) => {
+        if (err || !row) {
+          return cb(err);
+        }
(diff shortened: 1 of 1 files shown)


store/queries.js
```diff
@@ -383,6 +383,8 @@ function getPeers(db, input, player, cb) {
         return cb(err);
       }
       t.personaname = row.personaname;
+      t.name = row.name;
+      t.is_contributor = row.is_contributor;
```
howardchung: Doesn't this have to be read from the JSON file? @gu3st

store/queries.js
```diff
@@ -383,6 +383,8 @@ function getPeers(db, input, player, cb) {
         return cb(err);
       }
       t.personaname = row.personaname;
+      t.name = row.name;
```
howardchung: Also not sure name is defined, what is this the name of?

howardchung: name is the field we have for pro players' names, but it will not be in this query result unless we join with notable_players which is the pro players table.

store/queries.js
gu3st: Yes. IIRC I created a convenience function in utils

builder-247:
Oh god I'm butchering git, trying to fix this ;-;

store/queries.js
builder-247: Would this be the proper way to do it?
```js
db.first().from('players').leftJoin('notable_players', 'players.account_id', 'notable_players.account_id').where({
// etc.
```

howardchung: yes, I believe so. You can test on a local instance

PR #1839, opened 2019-01-28
Add player/records enpoint

builder-247:
This is to allow creating records overview on web, similar to dotabuff:
image

## TODO:
- [ ] Add rest of the fields to spec.js


builder-247:
Maybe each record should contain the full match object?

store/queries.js
```diff
@@ -376,18 +377,21 @@ function getPeers(db, input, player, cb) {
   // limit to 200 max players
   teammatesArr = teammatesArr.slice(0, 200);
   return async.each(teammatesArr, (t, cb) => {
-    db.first().from('players').where({
+    db.first().from('players').leftJoin('notable_players', 'players.account_id', 'notable_players.account_id').where({
```
howardchung: this needs to be fixed to return the correct data (or remove this change)

store/queries.js
```diff
@@ -451,6 +455,29 @@ function getMatchRankTier(match, cb) {
   });
 }
 
+function getPlayerRecords(player, cb) {
```
howardchung: this is going to have to be cached in some way. . . otherwise this one call is making 20+ calls to the database

store/queries.js
```diff
@@ -376,18 +377,24 @@ function getPeers(db, input, player, cb) {
   // limit to 200 max players
   teammatesArr = teammatesArr.slice(0, 200);
   return async.each(teammatesArr, (t, cb) => {
-    db.first().from('players').where({
-      account_id: t.account_id,
-    }).asCallback((err, row) => {
-      if (err || !row) {
+    db.first('players.account_id', 'personaname', 'name', 'avatar', 'avatarfull')
```
howardchung: you also need last_login I think

howardchung: also I suggest making this fix in a separate PR, otherwise it will get mixed up with the other changes.

store/queries.js
builder-247: Perhaps I should only call the getPlayerMatches function once and then sort the data from there? Or do you have other ideas how this should be handled?

howardchung: Yeah, that might be more efficient.

PR #1881, opened 2019-03-20
Added repatch service

builder-247:


svc/repatch.js
```diff
@@ -0,0 +1,34 @@
...
+        {
+          match_id: match.match_id,
+        }, cb,
+      );
+    }, (err) => {
+      process.exit(Number(err));
+    });
+  });
+}
+invokeInterval(rePatch, 12 * 60 * 60 * 1000);
```
howardchung: 12 hours instead of 24?

builder-247: Woops. This is what you get from copy pasting and not thinking it through. I'll fix it.

PR #1900, opened 2019-04-26
Added OR condition on heroStats query to include heroes with 0 result…

lancebendo:
Issue: https://github.com/odota/web/issues/1887

Modified the heroStats query to included heroes that have 0 games played (e.g. new heroes). It will then include the hero_id, pro_win, pro_pick and pro_ban attributes in GET /heroStats for those heroes.

### DIFF
--- a/svc/herostats.js
+++ b/svc/herostats.js
@@ -15,21 +15,21 @@ function doHeroStats(cb) {
       db.raw(`
               SELECT
               floor(avg_rank_tier / 10) as rank_tier,
-              sum(case when radiant_win = (player_slot < 128) then 1 else 0 end) as win, 
-              count(*) as pick,
-              hero_id 
+              sum(case when radiant_win = (player_slot < 128) then 1 when hero_id is null then null else 0 end) as win, 
+              case when hero_id is not null then count(*) else null end as pick,
+              heroes.id as hero_id
               FROM public_player_matches 
-              JOIN 
+              RIGHT JOIN heroes on heroes.id = public_player_matches.hero_id
+              LEFT JOIN 
               (SELECT * FROM public_matches
               TABLESAMPLE SYSTEM_ROWS(10000000)
-              WHERE ((start_time > ?
-              AND start_time < ?)
-              OR start_time IS NULL)
+              WHERE start_time > ?
+              AND start_time < ?
               AND avg_rank_tier IS NOT NULL)
               matches_list USING(match_id)
-              WHERE hero_id > 0
-              GROUP BY rank_tier, hero_id
-              ORDER BY hero_id
+              WHERE heroes.id > 0
+              GROUP BY rank_tier, hero_id, heroes.id
+              ORDER BY heroes.id
           `, [minTime, maxTime])
         .asCallback(cb);
     },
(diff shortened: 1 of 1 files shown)


howardchung:
I ran the new query with an old range and it didn't seem to pick up Mars (ID 129)?

howardchung:
Actually seems like this works for the 2nd and 3rd queries because they join from heroes, but not the first.

lancebendo:
Oops, my bad! I skipped testing the 1st query since I assumed it was just the same as the 2nd and 3rd. I'll do it right this time

lancebendo:
> I ran the new query with an old range and it didn't seem to pick up Mars (ID 129)?

I'm thinking of leaving the rank_tier, win and pick NULL for heroes with 0 matches. Since win and pick depends on rank_tier and it makes sense if rank_tier will be null in case a hero was never played yet. Any thoughts before I commit the changes?

howardchung:
That sounds fine. We'll probably have to check to see if the web UI handles that properly (I expect it should)

lancebendo:
Follow-up commit made for the 1st query.

svc/herostats.js
```diff
@@ -15,20 +15,21 @@ function doHeroStats(cb) {
       db.raw(`
               SELECT
               floor(avg_rank_tier / 10) as rank_tier,
-              sum(case when radiant_win = (player_slot < 128) then 1 else 0 end) as win, 
-              count(*) as pick,
-              hero_id 
+              sum(case when radiant_win = (player_slot < 128) then 1 when hero_id is null then null else 0 end) as win,
```
howardchung: i wonder if it would be cleaner to start with heroes and left join the other tables like the other two queries.

lancebendo: > i wonder if it would be cleaner to start with heroes and left join the other tables like the other two queries.

I agree. Okay, I'll refactor it.

svc/herostats.js
```diff
@@ -15,20 +15,21 @@ function doHeroStats(cb) {
...
               TABLESAMPLE SYSTEM_ROWS(10000000)
               WHERE start_time > ?
               AND start_time < ?
               AND avg_rank_tier IS NOT NULL)
               matches_list USING(match_id)
-              WHERE hero_id > 0
-              GROUP BY rank_tier, hero_id
-              ORDER BY hero_id
+              WHERE heroes.id > 0
+              GROUP BY rank_tier, hero_id, heroes.id
```
howardchung: hero_id not necessary if already grouping by heroes.id?

lancebendo: It's still necessary for the calculations of pick and win column. I only included the heroes table to include the heroes that are not in the public_player_matches table so I used heroes.id for hero_id column instead of public_player_matches.hero_id.

I'm thinking of changing the hero_id to public_player_matches.hero_id to at least make it more explicit.

lancebendo: > hero_id not necessary if already grouping by heroes.id?

It's still necessary for the calculations of pick and win column. I only included the heroes table to include the heroes that are not in the public_player_matches table so I used heroes.id for hero_id column instead of public_player_matches.hero_id.

I'm thinking of changing the hero_id to public_player_matches.hero_id to at least make it more explicit.

howardchung: Sure, that'd make it more clear.

PR #2026, opened 2019-12-28
updated method for getting client's ip address

builder-247:
* This should fix client's with IPv6 address being rate limited based on the last hexadectet of the address


howardchung:
conflicts from a dependency update earlier

builder-247:
ah, my archnemesis package-lock.json and git

Issue #1907, opened 2019-05-03
store parsed matches in ParseHistory table


howardchung:
A common request is to get IDs of only matches that are parsed. Currently we don't index those, but we can create a new table and store the IDs there.

dota2heqiuzhi:
> A common request is to get IDs of only matches that are parsed. Currently we don't index those, but we can create a new table and store the IDs there.

@howardchung Is it done？

howardchung:
Not yet. public_player_matches is a sample, so we'd probably want a new table.

dota2heqiuzhi:
Is there a schedule? This is very useful！

Howard Chung <notifications@github.com> 于2019年12月10日周二 下午2:47写道：

howardchung:
I don't have a set target date on this feature, but happy to accept a PR for it.

dota2heqiuzhi:
My English is not very good，so is my python.
I don't know if I can do it well

Howard Chung <notifications@github.com> 于2019年12月12日周四 上午8:03写道：

ahi-el:
hi @howardchung, just to make sure 

what columns should the table include? only matchId? do we want to return -all- the parsed matchesIds whenever the api is called?

howardchung:
Yes, I think matchID is sufficient.

To restrict the response size, we should probably limit to 100 per call, but allow paging back with less_than_match_id

PR #2027, opened 2019-12-28
create new parsed_matches table

ahi-el:
Creating a new parsed_matches in order to store all parsed matches
This is for https://github.com/odota/core/issues/1907

First PR, a new table
Next PR will be for the API call

### DIFF
--- a/store/queries.js
+++ b/store/queries.js
@@ -868,6 +868,17 @@ function insertMatch(match, options, cb) {
           });
       }
 
+      function upsertParsedMatch(cb) {
+        if (match.start_time) {
+          return upsert(trx, 'parsed_matches', {
+            match_id: match.match_id,
+          }, {
+            match_id: match.match_id,
+          }, cb);
+        }
+        return cb();
+      }
+
       function exit(err) {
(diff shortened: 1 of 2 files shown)


store/queries.js
```diff
@@ -868,6 +868,17 @@ function insertMatch(match, options, cb) {
           });
       }
 
+      function upsertParsedMatch(cb) {
+        if (match.start_time) {
```
howardchung: Why do we check for start_time here?

howardchung: I think if you only want to list parsed matches, you should check for match.version here

ahi-el: I went by another function
```
      function upsertMatchPatch(cb) {
        if (match.start_time) {
          return upsert(trx, 'match_patch', {
            match_id: match.match_id,
            patch: constants.patch[utility.getPatchIndex(match.start_time)].name,
          }, {
            match_id: match.match_id,
          }, cb);
        }
        return cb();
      }
```
but I see now that it verified match.start_time because it used it for the patch. I'll change it to match.version

PR #2037, opened 2019-12-30
Get parsed matches

### LINKED ISSUE #1907: store parsed matches in ParseHistory table

ahi-el:
fixes #1907

added a url path to get parsed matches, limited by 100 & using less_than_match_id

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -2328,8 +2328,8 @@ You can find data that can be used to convert hero and ability IDs and other inf
     '/parsedMatches': {
       get: {
         summary: 'GET /parsedMatches',
-        description: 'Get list of parsed matches',
-        tags: ['parsed public matches'],
+        description: 'Get list of parsed match IDs',
+        tags: ['parsed matches'],
         parameters: [
           params.lessThanMatchIdParam,
         ],
@@ -2356,7 +2356,7 @@ You can find data that can be used to convert hero and ability IDs and other inf
 
           db.raw(`
           SELECT * FROM parsed_matches
-          WHERE match_id <= ?
+          WHERE match_id < ?
           ORDER BY match_id DESC
           LIMIT 100
           `, [lessThan])
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -2325,6 +2325,49 @@ You can find data that can be used to convert hero and ability IDs and other inf
         },
       },
     },
+    '/parsedMatches': {
+      get: {
+        summary: 'GET /parsedMatches',
+        description: 'Get list of parsed matches',
+        tags: ['parsed public matches'],
+        parameters: [
+          params.lessThanMatchIdParam,
+        ],
+        responses: {
+          200: {
+            description: 'Success',
+            schema: {
+              type: 'array',
+              items: {
+                type: 'object',
+                properties: {
+                  match_id: {
+                    description: 'match_id',
+                    type: 'integer',
+                  },
+                },
+              },
+            },
+          },
+        },
+        route: () => '/parsedMatches',
+        func: (req, res, cb) => {
+          const lessThan = Number(req.query.less_than_match_id) || Number.MAX_SAFE_INTEGER;
+
+          db.raw(`
+          SELECT * FROM parsed_matches
+          WHERE match_id < ?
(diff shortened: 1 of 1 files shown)


routes/spec.js
```diff
@@ -2325,6 +2325,49 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+            },
+          },
+        },
+        route: () => '/parsedMatches',
+        func: (req, res, cb) => {
+          const lessThan = Number(req.query.less_than_match_id) || Number.MAX_SAFE_INTEGER;
+
+          db.raw(`
+          SELECT * FROM parsed_matches
+          WHERE match_id < ?
```
Argn0: in my opinion less than or equal would be  more intuitive

routes/spec.js
```diff
@@ -2325,6 +2325,49 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+        responses: {
+          200: {
+            description: 'Success',
+            schema: {
+              type: 'array',
+              items: {
+                type: 'object',
+                properties: {
+                  match_id: {
+                    description: 'match_id',
```
Argn0: I find this description too inaccurate

routes/spec.js
Argn0: probably doesn't matter though

routes/spec.js
ahi-el: What would you say is inaccurate about the description? match_id is pretty self explanatory.

Should it be something like "match id of a previously parsed match"?

routes/spec.js
ahi-el: sounds good

routes/spec.js
Argn0: nevermind. I have mistaken this for the query paremeter.

routes/spec.js
```diff
@@ -2325,6 +2325,50 @@ You can find data that can be used to convert hero and ability IDs and other inf
         },
       },
     },
+    '/parsedMatches': {
+      get: {
+        summary: 'GET /parsedMatches',
+        description: 'Get list of parsed matches',
```
howardchung: I'd change this to "match IDs"

routes/spec.js
```diff
@@ -2325,6 +2325,50 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+        tags: ['parsed public matches'],
```
howardchung: "parsed matches", this will include all

routes/spec.js
```diff
@@ -2325,6 +2325,50 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+                    type: 'integer',
+                  },
+                },
+              },
+            },
+          },
+        },
+        route: () => '/parsedMatches',
+        func: (req, res, cb) => {
+          const lessThan = Number(req.query.less_than_match_id) || Number.MAX_SAFE_INTEGER;
```
howardchung: I am pretty sure you can just pass this as a string to postgres and let the database handle it.

routes/spec.js
```diff
@@ -2325,6 +2325,50 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+
+          db.raw(`
+          SELECT * FROM parsed_matches
+          WHERE match_id <= ?
```
howardchung: should be <

routes/spec.js
howardchung: I think we should use < for consistency with the other method and the name

routes/spec.js
ahi-el: done

routes/spec.js
ahi-el: done

routes/spec.js
ahi-el: done

routes/spec.js
ahi-el: I'm not sure what you mean by this. Can you give an example?

howardchung: you can just remove the Number() cast

ahi-el: ah, got it now. Done

PR #2089, opened 2020-02-27
Added `/constants` endpoint

builder-247:


routes/spec.js
```diff
@@ -4343,6 +4343,28 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+          required: true,
+          type: 'string',
+        }],
+        route: () => '/constants/:resource',
+        func: (req, res, cb) => {
+          const { resource } = req.params;
+          if (resource in constants) {
+            return res.json(constants[resource]);
+          }
+          return cb();
```
howardchung: does this go to the 404 handler?

routes/spec.js
```diff
@@ -4343,6 +4343,28 @@ You can find data that can be used to convert hero and ability IDs and other inf
         },
       },
     },
+    '/constants/{resource}': {
```
howardchung: My worry with this name is that people won't realize this means static game data like hero stats etc.

Maybe we can add some keywords

builder-247: Any ideas?

routes/spec.js
```diff
@@ -4343,6 +4343,28 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+      get: {
+        summary: 'GET /constants',
+        description: 'Get constant data mirrored from the dotaconstants repository',
```
howardchung: maybe "static game data" instead of constant data

howardchung:
Tests failing? My guess is it's trying and failing to test the new endpoint.

builder-247:
All  builds seem to be failing for some reason during setup

PR #2092, opened 2020-03-03
remove tracking ip in api key usage code to avoid ipv6 issues

howardchung:


albertcui:
LGTM, do you know why the test is failing?

howardchung:
All the tests started failing a couple weeks ago due to it not being able to connect to postgres. I'm looking at it

PR #2101, opened 2020-03-12
Return list of resources if none is specified

builder-247:


routes/spec.js
```diff
@@ -4346,17 +4346,20 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
           description: 'Resource name e.g. `heroes`. [List of resources](https://github.com/odota/dotaconstants/tree/master/build)',
-          required: true,
+          required: false,
           type: 'string',
         }],
-        route: () => '/constants/:resource',
+        route: () => '/constants/:resource?',
         func: (req, res, cb) => {
+          if (!req.params.resource) {
+            return res.json({ resources: Object.keys(constants) });
```
howardchung: maybe this should just be the array directly?

PR #2116, opened 2020-03-27
Item popularity endpoint

ccyccyccy:
Aggregate item popularity for a specific hero, categorised into start of game, early game, mid game and late game items.

For use in [this feature](https://github.com/odota/web/issues/2478)

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -59,7 +59,7 @@ function sendDataWithCache(req, res, data, key) {
     cacheFunctions.write({
       key,
       account_id: req.params.account_id,
-    }, JSON.stringify(data), () => { });
+    }, JSON.stringify(data), () => {});
   }
   return res.json(data);
 }
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -59,7 +59,7 @@ function sendDataWithCache(req, res, data, key) {
     cacheFunctions.write({
       key,
       account_id: req.params.account_id,
-    }, JSON.stringify(data), () => {});
+    }, JSON.stringify(data), () => { });
   }
   return res.json(data);
 }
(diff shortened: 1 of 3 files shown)


store/queries.js
```diff
@@ -216,6 +216,55 @@ function getHeroRankings(db, redis, heroId, options, cb) {
   });
 }
 
+async function getItemFromName(db, redis, itemName) {
+  return db.raw(`
```
howardchung: I believe you can do this by referencing dotaconstants which would save a db call.

store/queries.js
```diff
@@ -216,6 +216,55 @@ function getHeroRankings(db, redis, heroId, options, cb) {
...
+  FROM items
+  WHERE name = ?
+  `, [itemName || 0]);
+}
+
+function getHeroItemPopularity(db, redis, heroId, options, cb) {
+  db.raw(`
+  SELECT purchase_log
+  from player_matches
+  JOIN parsed_matches USING(match_id)
```
howardchung: player_matches will only contain pro matches which are a subset of parsed matches, so I think this join is unnecessary

routes/spec.js
```diff
@@ -3710,6 +3710,74 @@ You can find data that can be used to convert hero and ability IDs and other inf
         },
       },
     },
+    '/heroes/{hero_id}/itemPopularity': {
+      get: {
+        summary: 'GET /heroes/{hero_id}/itemPopularity',
+        description: 'Get item popularity of hero categoried by start, early, mid and late game',
```
howardchung: might want to highlight this is only for pro games

store/queries.js
```diff
@@ -216,6 +216,55 @@ function getHeroRankings(db, redis, heroId, options, cb) {
...
+      const itemRow = (await getItemFromName(db, redis, itemName)).rows[0];
+      const cost = parseInt(itemRow.cost, 10);
+      const itemId = itemRow.id;
+      return {
+        itemId,
+        time,
+        cost,
+      };
+    }))
+      .then((items) => {
```
howardchung: can we use async/await instead?

store/queries.js
ccyccyccy: Oh didn't know about that. Fixed.

store/queries.js
ccyccyccy: Thanks, didn't know about this as well. Is there a place where these information are documented?

routes/spec.js
ccyccyccy: fixed

store/queries.js
ccyccyccy: No more async calls needed since there is no need to do db calls for getting item costs and id.

store/queries.js
ccyccyccy: Hmm I tried to delete the line locally, but the aggregation step crashed due to there being some entries that were not parsed. Is it just an issue on my local install?

howardchung: You can try querying for "version IS NOT NULL" to avoid picking up unparsed matches

ccyccyccy: Thanks that fixed the problem. But wouldn't joining with parsed_matches do the same thing?

howardchung: I believe parsed_matches is a list of all parsed matches, not just pro games

PR #2145, opened 2020-04-26
Add dotaplus xp

### LINKED ISSUE #1864: Add Dota Plus hero level tracking
- Update [odota/parser](https://github.com/odota/parser) to extract them from replays
- Save the data to a new Cassandra field
- Return the data from `/players/:account_id` endpoint

Also possibly store the level at the time of the match to the match data?

builder-247:
Closes #1864

### Changes

This PR adds `hero_dotaplus_xp` field to both `/player/matches/{match}` and `/player/heroes` endpoints

### DIFF
--- a/processors/createParsedDataBlob.js
+++ b/processors/createParsedDataBlob.js
@@ -21,6 +21,7 @@ function getParseSchema() {
     radiant_gold_adv: [],
     radiant_xp_adv: [],
     cosmetics: {},
+    dotaplus: {},
     players: Array(...new Array(10)).map(() => ({
       player_slot: 0,
       obs_placed: 0,
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1487,10 +1487,6 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     description: 'against_win',
                     type: 'integer',
                   },
-                  hero_dotaplus_xp: {
-                    description: 'Dota Plus hero experience',
-                    type: 'integer',
-                  },
                 },
               },
             },
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -472,6 +472,10 @@ You can find data that can be used to convert hero and ability IDs and other inf
                         description: 'The ID value of the hero played',
                         type: 'integer',
                       },
+                      hero_dotaplus_xp: {
+                        description: 'Dota Plus hero xp',
+                        type: 'integer',
+                      },
                       item_0: {
                         description: 'Item in the player\'s first slot',
                         type: 'integer',
@@ -1483,6 +1487,10 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     description: 'against_win',
                     type: 'integer',
                   },
+                  hero_dotaplus_xp: {
+                    description: 'Dota Plus hero experience',
+                    type: 'integer',
+                  },
                 },
               },
             },
@@ -1502,6 +1510,7 @@ You can find data that can be used to convert hero and ability IDs and other inf
               with_win: 0,
               against_games: 0,
               against_win: 0,
+              hero_dotaplus_xp: 0,
             };
             heroes[heroId] = hero;
           });
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1487,6 +1487,10 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     description: 'against_win',
                     type: 'integer',
                   },
+                  hero_dotaplus_xp: {
+                    description: 'hero_dotaplus_xp',
+                    type: 'integer',
+                  },
                 },
               },
             },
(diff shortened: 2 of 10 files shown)


builder-247:
The `dotaplus` object from the parsed replay doesn't seem to appear in `insertMatch` function, can anyone spot what I'm missing?

routes/spec.js
```diff
@@ -472,6 +472,10 @@ You can find data that can be used to convert hero and ability IDs and other inf
                         description: 'The ID value of the hero played',
                         type: 'integer',
                       },
+                      hero_dotaplus_xp: {
```
howardchung, line 475: Would this make more sense stored as a separate table from each player_hero combination? Seems like this would be a properly that's independent of each match, unless we care about their dota plus level at the time of the match.

sql/migrations.cql
```diff
@@ -1,2 +1,3 @@
 ALTER TABLE player_matches ADD damage_targets text;
 ALTER TABLE player_matches ADD connection_log text;
+ALTER TABLE player_matches ADD dotaplus text;
```
howardchung, line 3: You probably meant to add this to matches not player_matches, same for sql file

store/buildMatch.js
```diff
@@ -51,6 +51,7 @@ async function extendPlayerData(player, match) {
     cluster: match.cluster,
     lobby_type: match.lobby_type,
     game_mode: match.game_mode,
+    hero_dotaplus_xp: (match.dotaplus || {})[player.account_id] || 0,
```
howardchung, line 54: it looks like you're storing the dotaplus object in matches not player_matches, which is probably why you're not seeing data here

routes/spec.js
builder-247: This is how the game client displays it, also not sure if gets set to 0 on the replays if the player doesn't have a valid subscription anymore

store/buildMatch.js
builder-247: I fixed it and ran `docker system prune` but it still isn't showing up.

howardchung: Did you run the actual schema update on the databases, or did you try to completely purge the containers and let Docker re-create them?

routes/spec.js
```diff
@@ -1483,6 +1487,10 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     description: 'against_win',
                     type: 'integer',
                   },
+                  hero_dotaplus_xp: {
```
howardchung, line 1490: if this is going to be stored per-match I don't think it'll be in this response

store/buildMatch.js
builder-247: Complete purge.

howardchung: Can you go into the container and check if the columns are in the database table?
The JS code will remove any fields that aren't in the schema at insertion time.

builder-247: Looks like the new columns aren't in the table. I added them manually and found some other issues I need to figure out.

builder-247:
I'm going to add the data to more endpoints that requires more research on the codebase.

routes/spec.js
```diff
@@ -1502,10 +1510,11 @@ You can find data that can be used to convert hero and ability IDs and other inf
               with_win: 0,
               against_games: 0,
               against_win: 0,
+              hero_dotaplus_xp: 0,
             };
             heroes[heroId] = hero;
           });
-          req.queryObj.project = req.queryObj.project.concat('heroes', 'account_id', 'start_time', 'player_slot', 'radiant_win');
+          req.queryObj.project = req.queryObj.project.concat('heroes', 'account_id', 'start_time', 'player_slot', 'radiant_win', 'hero_dotaplus_xp');
```
howardchung, line 1517: I don't think we should make this a default returned field? Clients can request it specifically

sql/create_tables.cql
```diff
@@ -150,6 +151,7 @@ CREATE TABLE IF NOT EXISTS player_caches (
   gold_per_min text,
   xp_per_min text,
   hero_id text,
+  hero_dotaplus_xp text,
```
howardchung, line 154: we only need it in the table if we are going to aggregate this across matches for a player

routes/spec.js
```diff
@@ -1534,6 +1543,8 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     heroes[tmHero].against_games += 1;
                     heroes[tmHero].against_win += playerWin ? 1 : 0;
                   }
+                  // Use the highest xp value
+                  heroes[tmHero].hero_dotaplus_xp = Math.max(m.hero_dotaplus_xp, heroes[tmHero].hero_dotaplus_xp);
```
howardchung, line 1547: don't think this is necessary, isn't heroes just the object that tracks the heroes played with?

store/buildMatch.js
```diff
@@ -98,6 +101,7 @@ async function getMatch(matchId) {
   const gcdataPromise = db.first().from('match_gcdata').where({
     match_id: matchId,
   });
+  delete match.dotaplus;
```
howardchung, line 104: if this is now anonymized using player slot we don't have to delete?

routes/spec.js
howardchung: oh i see, looks like you're adding it here to display the hero level for all of a player's heroes

routes/spec.js
builder-247: Fair enough

processors/createParsedDataBlob.js
```diff
@@ -21,6 +21,7 @@ function getParseSchema() {
     radiant_gold_adv: [],
     radiant_xp_adv: [],
     cosmetics: {},
+    dotaplus: {},
```
howardchung, line 24: Maybe we should store this as a field per player instead of per match? since we need to store it per player in the player_caches table anyway

Issue #2194, opened 2020-07-03
Word cloud for different languages


MVsevolod:
This will be a nice feature to parse not only english, but also other languages for word cloud statistics.

howardchung:
I think we restricted it because if we add e.g. chinese characters, the number of words goes way up and the word cloud loses a lot of meaning, so we currently filter to only words with [a-z] letters.

Maybe we could include cyrillic letters to allow languages like Russian, though.

howardchung:
For reference this is the function converting a chat message line into words:
https://github.com/odota/core/blob/master/util/utility.js#L23

PR #2234, opened 2020-09-11
Add wordcloud support for cyrillic letters (#2194)

builder-247:


util/utility.js
```diff
@@ -21,7 +21,7 @@ const contributors = require('../CONTRIBUTORS');
  * @return {Array}
  */
 function tokenize(input) {
-  return input.replace(/[^a-zA-Z- ]+/g, '').replace('/ {2,}/', ' ').toLowerCase().split(' ');
+  return input.replace(/[^a-zwа-я- ]+/gi, '').replace('/ {2,}/', ' ').toLowerCase().split(' ');
```
howardchung, line 24: Can you explain how this works? I tossed it into https://regex101.com/ and seems like it matches a-z, then w, then the cyrillic letter set? What about capitals?

builder-247: Looks like the `w` got in by mistake. The `i` flag takes capitalization into account.

howardchung (approved):

PR #2254, opened 2020-10-14
Added starting items to /matches endpoint

27bslash:
odota/parser#42

### DIFF
--- a/processors/processLogParse.js
+++ b/processors/processLogParse.js
@@ -6,13 +6,12 @@ const insignificantDeaths = [
 function translate(s) {
   return s === 'dota_unknown' ? null : s;
 }
-
 /**
  * A processor to reduce the event stream to only logs we want to persist
  * */
 function processReduce(entries, meta) {
   const result = entries.filter((e) => {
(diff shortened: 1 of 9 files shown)


processors/processLogParse.js
```diff
@@ -6,13 +6,12 @@ const insignificantDeaths = [
...
   return s === 'dota_unknown' ? null : s;
 }
-
 /**
  * A processor to reduce the event stream to only logs we want to persist
  * */
 function processReduce(entries, meta) {
   const result = entries.filter((e) => {
-    if (e.type === 'DOTA_COMBATLOG_PURCHASE'
+    if (e.type === 'DOTA_COMBATLOG_PURCHASE' || e.type === 'STARTING_ITEM'
```
howardchung, line 14: I think we can leave this out, we don't need this in the pro game logs.

howardchung (approved):

howardchung:
This looks generally fine, but we might want to optimize how we store this data, as we'd be storing this x10 players for each match and storage is already the most expensive part of running the service.

PR #2262, opened 2020-10-26
Fixed items with charges not being added correctly to purchase log

27bslash:
Currently Items that can be stacked do not show up correctly in the items tab:
image
image

Changed to add copies of the item depending on the max charges and the current charges an item has.
image
image

### DIFF
--- a/processors/populate.js
+++ b/processors/populate.js
@@ -62,6 +62,13 @@ function populate(e, container) {
             time: e.time,
             key: e.key,
           };
+          if (e.type === 'purchase_log' && e.charges) {
+            let maxCharges = 1;
+            if (e.key === 'tango') maxCharges = 3;
+            for (let i = 1; i < maxCharges / e.charges; i += 1) {
+              t.push(arrEntry);
+            }
+          }
           if (e.type === 'kills_log' && e.tracked_death) {
             arrEntry = Object.assign({}, {
               tracked_death: e.tracked_death,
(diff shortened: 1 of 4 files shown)


processors/populate.js
```diff
@@ -62,6 +62,13 @@ function populate(e, container) {
             time: e.time,
             key: e.key,
           };
+          if (e.type === 'purchase_log' && e.charges) {
+            let maxCharges = 1;
+            if (e.key === 'tango') maxCharges = 3;
+            for (let i = 1; i < maxCharges / e.charges; i += 1) {
```
howardchung, line 68: Why not just push the entry with the charges field? We can show a number on the UI to indicate number of charges

howardchung:
I think we should just show a number rather than duplicating the item

processors/populate.js
27bslash: I thought about it, but I thought that the charge number would be too small and make the item image too busy.

27bslash:
Would something like this be fine then?
image
and if it is should I make a new pull  request or update this one?

howardchung:
You can update this one. Maybe make the number a bit smaller and closer to the top right corner.

27bslash:
Not sure what the etiquette here is. But I have made the changes you requested. https://github.com/odota/web/pull/2685 

Follow up question though, do you have any ideas as to why the travis ci build is failing. It says `cannot find module aws-sign2` but the files changed shouldn't effect this.

howardchung:
Thanks for the update, I'll review this when I have some time!

That build failure is probably affecting all builds, I'll get it fixed and then it should go away but shouldn't block this

PR #2295, opened 2020-12-08
Update to ingest new DOTA_ABILITY_LEVEL

slavox:
Update bountyhunter & alchemist gold values
Re-enable performanceOthers

This pull request also requires the parser update

### DIFF
--- a/processors/createParsedDataBlob.js
+++ b/processors/createParsedDataBlob.js
@@ -62,6 +62,7 @@ function getParseSchema() {
       xp_reasons: {},
       killed: {},
       item_uses: {},
+      ability_levels: {},
       ability_uses: {},
       ability_targets: {},
       damage_targets: {},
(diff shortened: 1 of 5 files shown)


processors/createParsedDataBlob.js
```diff
@@ -62,6 +62,7 @@ function getParseSchema() {
       xp_reasons: {},
       killed: {},
       item_uses: {},
+      ability_levels: {},
```
howardchung, line 65: If you only need this for checking ability level and don't plan to save it to the database, I'd just stick this in the meta variable rather than this

slavox: I have updated the code as requested. 
I just wanted to check is it okay to mutate meta inside populate.js? I didn't immediately see any other cases of this but in the case of abilities we have to gather them during the course of the loop.

slavox:
Is there anything else I should change?

howardchung (approved):

PR #2301, opened 2020-12-21
Changed data type for purchase_log key

### LINKED ISSUE #2300: Swagger api spec issue - purchase log key data type
**Issue with routes/spec.js** (line 625? Permalink below)
https://github.com/odota/core/blob/bcb1b4808fe6d5bf2675d89c01b3428ea2aa1c04/routes/spec.js#L625

Swagger spec says that the data type for the key on a purchase_log element is "integer", but responses coming back from a match query seem to be strings. Causes error on deserialization.

Sample response:
`'purchase_log': [{'charges': None,
                                'key': 'orb_of_venom',
                                'time': -89},`

Schoolboy215:
Key was previously integer, but the server responses are strings. Causes deserialization error

fixes #2300

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -621,8 +621,8 @@ You can find data that can be used to convert hero and ability IDs and other inf
                               type: 'integer',
                             },
                             key: {
-                              description: 'Integer item ID',
-                              type: 'integer',
+                              description: 'String item ID',
+                              type: 'string',
                             },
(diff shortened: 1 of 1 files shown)


Schoolboy215:
This PR is for issue 2300

https://github.com/odota/core/issues/2300

howardchung (approved):

Issue #1703, opened 2018-08-09
Swagger Spec Validation Errors


saip92:
I don't think this blocks anyone but I just wanted to point out that when using `https://api.opendota.com/api` in [Swagger Editor](https://editor.swagger.io/), it throws a bunch of schema errors in the spec.

I'm not sure if it was once a valid spec but I see that the spec was manually composed [here](https://github.com/odota/core/blob/master/routes/spec.js) and is probably just not up to date?

howardchung:
Did the official Swagger spec change/release a new version?  I'm not sure what the correct fix is.  We do manually compose it, but it seems to work fine with the API docs: https://docs.opendota.com

Sheemap:
@howardchung I ran into this issue today, so its still around. I'm not sure why the doc generation works, but the spec doesn't appear to have changed, OpenDota's implementation just has always been invalid lol

I've scoured the [OpenApi 2.0 docs](https://swagger.io/specification/v2/), and here are the issues that cause the errors:

- GET /constants/{resource}
    - Path parameters must be required (Currently false)
    - The responses object is required. (Currently missing)
- dateTime type
    - This type doesnt exist, and should actually be of "string" with a format of "date-time"
- GET /players/{account_id}/rankings
    - The "properties" field is misspelled as "properies"

Another note on the dateTime type, according to the spec the "date-time" format should follow [RFC3339](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14), but I've noticed a comment in the spec that describe the time sent back as `Time in seconds of last login of the player`, I assume thats an epoch timestamp, which doesnt follow that RFC return value. 

So in fixing the spec, we would need to make sure that the underlying endpoints return the correct data types we indicate to send back.

I'm happy to help work on a PR for this, and potentially implement a unit test that verifies the spec's accuracy to prevent it from happening again.

howardchung:
Sure, a PR fixing these issues would be very welcome!

For the date maybe we can just change the type to number

Sheemap:
Sounds good to me, Ill submit that soon!

On dateTime, Im moving through some of the fields and finding that they return it in the correct RFC format, looks like thats just how Postgres naturally returns it from the database when its stored as `timestamp with time zone`. So Ill go through and change it to that, or integer depending on what the code returns.

[...]

PR #2372, opened 2021-06-24
Resolve swagger parsing issues

Sheemap:
Resolves https://github.com/odota/core/issues/1703

This doesn't fix the incorrect API definitions, but does fix parsing. So tools can be used to generate clients that are mostly correct.

This PR also adds a new dependency, I'm not sure the guidelines on adding those, but it adds https://www.npmjs.com/package/@apidevtools/swagger-parser in order to parse the spec in a unit test to ensure validity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "url": "http://github.com/odota/core"
   },
   "dependencies": {
+    "@apidevtools/swagger-parser": "^10.0.2",
     "JSONStream": "^1.3.5",
     "async": "^2.6.1",
     "body-parser": "^1.19.0",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "url": "http://github.com/odota/core"
   },
   "dependencies": {
-    "@apidevtools/swagger-parser": "^10.0.2",
     "JSONStream": "^1.3.5",
     "async": "^2.6.1",
     "body-parser": "^1.19.0",
@@ -64,7 +63,8 @@
     "eslint-config-airbnb": "^17.1.0",
     "eslint-plugin-import": "^2.20.2",
     "mocha": "^8.2.1",
-    "prettier": "^2.0.5"
+    "prettier": "^2.0.5",
+    "@apidevtools/swagger-parser": "^10.0.2"
   },
   "engines": {
     "node": "12.14.0"
--- a/package.json
+++ b/package.json
@@ -59,12 +59,12 @@
     "uuid": "^3.3.3"
   },
   "devDependencies": {
+    "@apidevtools/swagger-parser": "^10.0.2",
     "eslint": "^5.16.0",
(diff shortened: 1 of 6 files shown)


routes/spec.js
```diff
@@ -4481,6 +4495,36 @@ You can find data that can be used to convert hero and ability IDs and other inf
...
+              type: 'array',
+              items: {
+                type: "string",
+              },
+            },
+          },
+        },
+        route: () => '/constants',
+        func: (req, res, cb) => {
+          if (!req.params.resource) {
```
howardchung, line 4517: this will never have the resource parameter?

package.json
```diff
@@ -25,6 +25,7 @@
     "url": "http://github.com/odota/core"
   },
   "dependencies": {
+    "@apidevtools/swagger-parser": "^10.0.2",
```
howardchung, line 28: could be a devdependency?

routes/spec.js
Sheemap: You're right. I cleared out the code that will never be hit

package.json
Sheemap: Good point, didn't even think about that. Moved

howardchung (approved):

howardchung:
unclear why CI is failing--maybe due to the package-lock change?

travis error is:
```
$ sudo bash scripts/build.sh
Error: Cannot perform an interactive login from a non TTY device
The command "sudo bash scripts/build.sh" exited with 1.
```

Sheemap:
I think its actually related to the DOCKER_PASSWORD environment var, its not getting set somehow.

I was able to replicated it locally by setting DOCKER_USERNAME, but leaving DOCKER_PASSWORD unset. With that, the docker login command within the build script fails with that same interactive login error message

Though Im not sure why the master build you recently tried succeeded fine. Is the environment variable dependent on who is running the build? Like I don't have access to the DOCKER_PASSWORD var, so it doesnt set it?

Sheemap:
Yeah, looking more into it, Im almost positive thats the issue.

Your master build output has this near the top

```
Setting environment variables from repository settings
$ export DOCKER_USERNAME=howardc93
$ export DOCKER_PASSWORD=[secure]
$ export STRIPE_SECRET=[secure]
$ export STRIPE_API_PLAN=[secure]
```

While mine only has this
```
Setting environment variables from repository settings
$ export DOCKER_USERNAME=howardc93
```

So for some unknown reason, this build isn't getting the password set.

howardchung:
Oh I believe this is due to security reasons, PRs don't get env vars.

Sheemap:
Ah interesting, I thought package.json was the only file that controlled versions.

I can create a new PR resolving those rules if you'd like. Looks like its a few really simple fixes (and I introduced a few of em : ) . I didn't check to see if linting was a part of the build process

PR #2374, opened 2021-06-25
Resolve linting issues

Sheemap:
This PR re-enables linting within the build command, and resolves all the errors the linter spit out before.


howardchung:
Thanks! Forcing lint as part of build helps enforce consistency, but is sometimes annoying when version differences cause the linter to complain even with non-blocking issues

PR #2428, opened 2021-10-02
Harmonize output between match list API endpoints

cschuijt:
I was working on a bot with a lastmatch command, for which we wanted to implement a hero filter. The output of the command includes basic info about someone's last game, like KDA and GPM/XPM. The original implementation used `/players/:player_id/recentMatches`, which returns all the relevant data but doesn't allow us to filter by hero. While I was switching it over to `/players/:player_id/matches` to be able to filter, I noticed some fields (GPM/XPM) are not returned here. On top of that, the documentation of `recentMatches` makes no mention of their inclusion there. This PR fixes both these things by making `/players/:player_id/matches` return the same fields per match as `/players/:player_id/recentMatches`, and making sure the documentation includes all fields being returned.

Since this pull request only adds fields to the API output, I don't think it's a breaking change, but I'm not familiar enough with the codebase to be sure about that. I also don't actually know whether there's a reason the output differed between these endpoints and my javascript is not super great, so if something in here doesn't make sense or could be done better, please let me know! I'm just trying to get on with my original project.

### DIFF
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1272,8 +1272,32 @@ You can find data that can be used to convert hero and ability IDs and other inf
                     description: 'Skill bracket assigned by Valve (Normal, High, Very High)',
                     type: 'integer',
                   },
+                  xp_per_min: {
+                    description: 'Experience per minute of the player',
+                    type: 'integer',
+                  },
+                  gold_per_min: {
+                    description: 'Gold per minute of the player',
+                    type: 'integer',
+                  },
+                  hero_damage: {
+                    description: 'Amount of damage dealt by the player to other heroes',
+                    type: 'integer',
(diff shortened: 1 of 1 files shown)


howardchung:
The intention here was that matches only returns a small set of fields, unless the "project" parameter is used to select fields to include.

I think we should update the recentMatches documentation though, that was just a copy/paste error.

cschuijt:
Sorry, I completely missed the project field in the docs! In hindsight, the part I was working on makes a lot more sense now. Is there a way to expand the docs to clarify what the project field is used for? It's just a single line at the moment, no way to tell what the (non-default) fields you can ask for are called, and it's easy to miss if no one points you in the right direction first.

PR #2492, opened 2022-02-10
API Billing Fixes

albertcui:
- disable keys when payment fails
- allow people to create new keys if previous payment fails ONLY if no outstanding balance


howardchung:
Can you split the formatting changes into their own PR?

routes/README_FOR_API_KEYS.md
```diff
@@ -0,0 +1,8 @@
+Make sure to set these ENV variables:
+
+STEAM_API_KEY=761FE7D69D5546AB1C053A911F14AF08
```
howardchung, line 3: Is this a real key?

svc/apiadmin.js
```diff
@@ -87,92 +87,91 @@ function storeUsageCounts(cursor, cb) {
...
+          `
+                  UPDATE api_keys SET is_canceled = true WHERE subscription_id = ?
+                `,
+          [sub.id]
+        );
 
-  if (cursor) {
-    options.starting_after = cursor;
-  }
+        continue;
```
howardchung, line 114: Are we letting the background job remove the canceled key from redis?

howardchung (approved):

howardchung:
If the user has an outstanding balance, do they have a way of paying it from our UI?

svc/apiadmin.js
albertcui: updateAPIKeysInRedis deletes all keys so just let that take care of it

PR #2571, opened 2022-07-01
Add match average rank

builder-247:
* To be used in player overview page

This should work right? It is kinda hard to test the scanner stuff - if I understood it correctly all new matches originate from scanner

### DIFF
--- a/sql/create_tables.cql
+++ b/sql/create_tables.cql
@@ -122,6 +122,7 @@ CREATE TABLE IF NOT EXISTS player_matches (
   firstblood_claimed int,
   party_size int,
   net_worth int,
+  average_rank int,
 );
 
 CREATE TABLE IF NOT EXISTS player_caches (
@@ -191,7 +192,8 @@ CREATE TABLE IF NOT EXISTS player_caches (
   item_3 text,
   item_4 text,
   item_5 text,
-  item_neutral text
+  item_neutral text,
+  average_rank int,
 );
 
 ALTER TABLE matches WITH compaction = { 'class' :  'LeveledCompactionStrategy'  };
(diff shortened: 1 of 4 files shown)


sql/create_tables.cql
```diff
@@ -122,6 +122,7 @@ CREATE TABLE IF NOT EXISTS player_matches (
   firstblood_claimed int,
   party_size int,
   net_worth int,
+  average_rank int,
```
howardchung, line 125: don't think it's needed here

sql/create_tables.sql
```diff
@@ -135,7 +135,8 @@ CREATE TABLE IF NOT EXISTS player_matches (
   roshans_killed int,
   observers_placed int,
   party_size int,
-  net_worth int
+  net_worth int,
+  average_rank int
```
howardchung, line 139: also not needed here since this is pro matches only

store/queries.js
```diff
@@ -485,7 +485,7 @@ function getMatchRankTier(match, cb) {
     }
     // Remove undefined/null values
     const filt = result.filter(r => r);
-    const avg = Math.floor(filt.map(r => Number(r)).reduce((a, b) => a + b, 0) / filt.length);
+    const avg = Math.floor(filt.map(r => Number(r)).reduce((a, b) => a + b, 0) / filt.length) || null;
```
howardchung, line 488: when will this be null?

builder-247: If none of the players have rank data

builder-247:
Should I also remove the `skill` field from the API response?

howardchung (approved):

PR #2574, opened 2022-07-05
Added pagination to teams endpoint

builder-247:
* This allows querying more than 1000 teams


howardchung (approved):
this is fine, although changes in rating between queries may lead to duplicate/missing results if someone is trying to fetch them all

PR #2582, opened 2022-07-12
fix average medal calculation

builder-247:
The old average rank function was made for mmr, so it didn't account for medal algebra, causing ranks such as Guardian 8 to appear.

### DIFF
--- a/util/utility.js
+++ b/util/utility.js
@@ -448,6 +448,20 @@ function average(data) {
 }
 
 /**
+ * Finds the average rank medal of input array
+ * */
+function averageMedal(values) {
+  let medalSum = 0;
+  let starSum = 0;
+  values.forEach((value) => {
+    medalSum += Number(String(value)[0]);
+    starSum += value % 10;
+  });
+  const len = values.length;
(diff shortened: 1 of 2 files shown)


util/utility.js
```diff
@@ -447,6 +447,20 @@ function average(data) {
...
+ * */
+function averageMedal(values) {
+  let medalSum = 0;
+  let starSum = 0;
+  values.forEach((value) => {
+    medalSum += Number(String(value)[0]);
+    starSum += value % 10;
+  });
+  const avgStars = (medalSum * 5 + starSum) / values.length;
+  return Math.floor(avgStars / 5) * 10 + Math.round(avgStars % 5);
```
howardchung, line 461: I assume this is basically converting all medals to a number of stars where 5 stars = 1 medal up, averaging the raw star number and then converting back?

howardchung (approved):

util/utility.js
howardchung: you could probably write this a little more simply by mapping the input array to a "numstars" array and then averaging that array before converting back. then you don't need the medalSum and starSum vars?

howardchung (approved):

PR #2583, opened 2022-07-15
Average medal always returns at  least 1 star

builder-247:


howardchung:
Wait how are we getting 0 here?

builder-247:
If avgStars is less than 0.5 mod 5

howardchung (approved):

PR #2609, opened 2022-10-01
Update courier death event

builder-247:
This PR fixes the incorrect team value in the `CHAT_MESSAGE_COURIER_LOST` lost event and adds killer and gold bounty information.

### Background

Before patch 7.23 (released 2019-11-26), each team only had one courier. The chat event would report `player1` property as the owner team of the courier, 2 for radiant and 3 for dire.

When each player got their own courier, the team property changed to `player2` with `player1` containing the slot of the killer, 0-4 for radiant and 5-9 for dire. For creep/tower killer the value is -1.

---
Related issues: https://github.com/odota/web/issues/3010, https://github.com/odota/web/issues/2778, https://github.com/odota/web/issues/2723, https://github.com/odota/web/issues/2443

### DIFF
--- a/processors/processExpand.js
+++ b/processors/processExpand.js
@@ -480,11 +480,20 @@ function processExpand(entries, meta) {
       });
     },
     CHAT_MESSAGE_COURIER_LOST(e) {
-      // player1 = team that lost courier? (2/3)
+      // For backwards compatability - when teams only had one courier player1 would be the team
+      let team;
+      if (e.player2 === undefined) {
+        team = e.player1;
+      } else {
+        // 2 is radiant and 3 dire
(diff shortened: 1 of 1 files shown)


processors/processExpand.js
```diff
@@ -480,11 +480,28 @@ function processExpand(entries, meta) {
       });
     },
     CHAT_MESSAGE_COURIER_LOST(e) {
-      // player1 = team that lost courier? (2/3)
+      let team;
```
howardchung, line 483: can we switch the behavior based on match ID?

howardchung: nvm looks like you use the existence of player2 to fix it

howardchung (approved):

PR #2653, opened 2023-04-21
Update for patch 7.33 Greevils Greed changes

slavox:
I don't currently play Dota so can't test this change, But assuming they kept the innate ability named the same this should work as before.

### DIFF
--- a/processors/performanceOthers.js
+++ b/processors/performanceOthers.js
@@ -4,27 +4,9 @@ function greevilsGreed(e, container, meta) {
     const alchSlot = meta.hero_to_slot[alchName];
     const alchPlayer = container.players[alchSlot];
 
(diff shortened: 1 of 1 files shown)


howardchung:
Ideally we'd handle both cases since we may reparse older matches. But I don't think this is a heavily used feature anyway so it might be fine

howardchung (approved):

slavox:
Can you point me to a place in the code where we can tell what patch is being parsed? I can update this to be safe for re-parsing

howardchung:
We usually calculate from the match's start_time properly using a utility function. I'm not sure if it's accessible in this processor though.

slavox:
Thanks, I found the function - Will investigate and create another pull request if I can get it working

PR #2654, opened 2023-04-27
Fix Swagger Spec definitions

ff137:
Very simple fix - in /benchmarks there are decimal values that are currently typed as integers.

Edit: Fixing more types. Log of changes:
- benchmark metrics from integer to number
- distributions: `dataTypeModifier` from string to integer
- findMatches: input types for teamA and teamB were integers, now array of integers. response type now correct as array
- leagueObject properties should be nullable
- matches required the most updates... summary:
 
- - player_slot property is nullable
- - objectives, picks_bans, gold_adv and xp_adv are arrays
- - expand cosmetics with item properties
- - additional units is an array and nullable
- - teamfights is a nested array
- - cosmetics here are integers (there are two fields named cosmetics in /matches response, just different nesting)
- - tpscrolls is an integer
- - last_login is nullable
- nullable fields in players response
- so many nullable fields
- rankings responds with an array; score is a number
- NB: RecordsResponse -- server responds with strings; should be fixed serverside and types reverted to integers, in order to be consistent with other match id responses
- NB: content type is `application/json; charset=utf-8`, previously just `application/json`
- `/constants/{resource}` response can be oneOf 3 different types
- managed to get all `/constants/{resource}` and `/players/{account_id}/{field}` paths to succeed

### DIFF
--- a/routes/objects.js
+++ b/routes/objects.js
@@ -240,14 +240,17 @@ const leagueObject = {
       ticket: {
         description: "ticket",
         type: "string",
+        nullable: "true",
       },
       banner: {
         description: "banner",
         type: "string",
+        nullable: "true",
       },
       tier: {
         description: "tier",
         type: "string",
+        nullable: "true",
       },
       name: {
         description: "name",
--- a/routes/properties.js
+++ b/routes/properties.js
@@ -2,6 +2,7 @@ module.exports = {
   radiant_win: {
     description: "Boolean indicating whether Radiant won the match",
     type: "boolean",
+    nullable: true,
   },
   player_slot: {
     description:
--- a/routes/properties.js
+++ b/routes/properties.js
@@ -7,6 +7,7 @@ module.exports = {
     description:
       "Which slot the player is in. 0-127 are Radiant, 128-255 are Dire",
     type: "integer",
+    nullable: true,
   },
   duration: {
     description: "Duration of the game in seconds",
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -2538,7 +2538,7 @@ You can find data that can be used to convert hero and ability IDs and other inf
                 properties: {
                   hero_id: {
                     description: "The ID value of the hero played",
-                    type: "string",
+                    type: "integer",
                   },
                   score: {
                     description: "hero_score",
--- a/routes/spec.js
+++ b/routes/spec.js
@@ -1153,23 +1153,26 @@ You can find data that can be used to convert hero and ability IDs and other inf
             description: "Success",
             schema: {
               title: "PlayersByRankResponse",
-              type: "object",
-              properties: {
-                account_id: {
-                  description: "account_id",
-                  type: "number",
+              type: "array",
+              items: {
+                type: "object",
+                properties: {
+                  account_id: {
+                    description: "account_id",
+                    type: "number",
+                  },
+                  rank_tier: {
+                    description:
+                      "Integer indicating the rank/medal of the player",
+                    type: "number",
+                  },
+                  fh_unavailable: {
+                    description:
+                      "Indicates if we were unable to fetch full history for this player due to privacy settings",
+                    type: "boolean",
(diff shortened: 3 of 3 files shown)


howardchung (approved):

ff137:
@howardchung I'm enjoying the reverse engineering, but this is turning out to be a whole lot more edits than I anticipated! Is there some way to generate the spec again?

ff137:
@howardchung NB: The server seems to return `match-id` as a String in RecordsResponse (`/records/<field>), but seems to be integer everywhere else. 

Meaning the spec looks right [here](https://github.com/odota/core/blob/40791bd5a9a1aa54ce3c6c1035e5339728723679/routes/spec.js#LL4765C1-L4766C1) but the server responds with string - so perhaps should be fixed elsewhere

howardchung:
Yeah the type should be fixed on the server in some spots.

ff137:
> Yeah the type should be fixed on the server in some spots.

So far it's most notably for `GET /records/{field}`, where the server response returns all strings when spec shows ints

Do you think that's something I can take a look at, or are you on top of it? I'm not sure where it is in the codebase

howardchung:
You can take a look if you like--I may not have time to get to it right now

ff137:
> You can take a look if you like--I may not have time to get to it right now

@howardchung Ok, where should I look I want to make those server-side changes? I can't find anything just by searching

Btw, I'm happy with the changes so far. 
- I took the changes that I made here (in the spec.js file), and applied it to the [opendota-api-spec](https://api.opendota.com/api);
- I converted the Swagger 2.0 spec to OpenAPI 3.x (with `swagger2openapi`), and a script to clean up some warnings
- and then finally generated a python opendota_client from the OpenAPI spec using `openapi-generator-cli generate`

I set up a very basic test to call `get` for almost all of the routes (didn't do `/search` or `/request`). 
The changes logged in the OP is what I did incrementally until all basic get methods run successfully. 

There is almost certainly more changes required - most likely more nullable fields in more rare cases - and other routes remain to be tested. But I think this is a good enough point to get reviewers involved, because I don't know how consequential these spec changes are to other systems. 

Any feedback will be appreciated

ff137 (commented):
I've reviewed all changes; formatting is now consistent with how it was

howardchung:
It looks like the test is complaining about an invalid spec--but I don't think it breaks the API functionality:
https://github.com/odota/core/actions/runs/4833780260/jobs/8614196665

ff137:
Ah yeah, I see the error in the test seems to be related to `GET /constants/{resource}`

And that route has the following new spec:
```
        responses: {
          200: {
            description: "Success",
            schema: {
              oneOf: [
                {
                  type: "object",
                  additionalProperties: {
                    title: "ConstantResourceResponse",
                  },
                },
                {
                  type: "array",
                  items: {
                    type: "object",
                    additionalProperties: {
                      title: "ConstantResourceResponse",
                    },
                  },
                },
                {
                  type: "array",
                  items: {
                    type: "integer",
                  },
                },
              ],
            },
          },
```

So, I found that `oneOf` combination allowed all the constant/resource endpoints to succeed.

But maybe the test is failing because 'null' should be one of the response types? A fix could be to add: 
```
oneOf: [
  // The other schemas...
  {
    type: "null",
  },
],
```

Or, it's because the properties needs to be defined when using `additionalProperties`:
```
          type: "object",
          additionalProperties: {
            title: "ConstantResourceResponse",
          },
          properties: {
            // Define the expected properties here
          },
        },
        {
          type: "array",
          items: {
            type: "object",
            additionalProperties: {
              title: "ConstantResourceResponse",
            },
            properties: {
              // Define the expected properties here
            },
          },
```

@howardchung I can check this out further, but idk how to run the Git Action tests without merging. Any suggestions?

howardchung:
You can run npm test locally, or submit a PR and I think I can get it to run the tests

PR #2656, opened 2023-04-28
Fix/pass test job w new spec

ff137:
Work in progress: 
- update spec to OpenAPI 3.0.3
- move model/parameter definitions to `components`
- ensure tests pass with new spec

Was in the process of updating CommonJS `require` pattern to ES6 imports, but this requires plenty of refactoring and validation that it works with dependencies. Can be done later


howardchung (approved):

ff137:
@howardchung I think the only way forward is to upgrade the spec to OpenAPI 3.1 and then to use OpenApiParser instead of SwaggerParser in the test. Busy setting that up, and I've pushed an example of what the openapi spec would look like. 

You can copy paste the json here to see how the OpenAPI spec renders: https://editor-next.swagger.io/

All that remains is for me to re-add the route/func definitions to the auto-generated conversion, and to use `components` to define repeated parameters and schemas. That last part will probably be the most work, going through every route and adding the re-used params/schemas to the list of components. But, if you're cool with the idea to upgrade to OpenAPI 3, then I'm happy to proceed and chip away at it over next few days or week

howardchung:
Looks like it's failing with a syntax error--es6 imports are not yet supported yet

howardchung:
Also, is it fine if I revert the previous merge and you can include those changes in this PR?

ff137:
> Also, is it fine if I revert the previous merge and you can include those changes in this PR?

Absolutely, I was surprised how quickly it was merged! I wanted to figure out about downstream impacts or what tests were necessary. Took a break since last week so I'll continue plugging away at it. Probably lots of small things to check so I'll mark PR ready when tests pass

> Looks like it's failing with a syntax error--es6 imports are not yet supported yet

Ah yeah I started updating the CommonJS `require` statements to ES6 imports. Kind of my own OCD to want things up to date if they can be, but seems very minor and nit-picky of me. It needs some other config changes in `package.json` to work properly. If there's any reason to stick to CommonJS then please let me know, otherwise I assume the upgrades are welcome

howardchung:
I would maybe make the syntax changes in a separate PR so it's less mixed together

ff137:
> I would maybe make the syntax changes in a separate PR so it's less mixed together

Indeed, decided the same!

I'm almost ready to push the latest changes for the spec. It looks ... much cleaner :-)

ff137:
@howardchung boom, the spec is valid. now moving on to tests.

notes:
- created responses.js to collect all response definitions. this can be better organised, but for now this made sense
- params + responses are neatly defined in components
- a few more params can be ref'd out of the spec, responses can also be refactored to use common types
- added a couple to-do's

ff137:
Excellent, the build test is passing now.

Some more work to do before I'm completely happy with it:
E.g. `picks_bans` in MatchResponse has the follow definition:
  ```json
      picks_bans: {
        description:
          "Object containing information on the draft. Each pick/ban contains a boolean relating to whether the choice is a pick or a ban, the hero ID, the team the picked or banned it, and the order.",
        type: "array",
        items: {
          type: "object",
        },
      },
```
In order to use an openapi client generator, these objects need to be defined:
```
picks_bans: {
  description: "Array containing information on the draft. Each item contains a boolean relating to whether the choice is a pick or a ban, the hero ID, the team the picked or banned it, and the order.",
  type: "array",
  items: {
    type: "object",
    properties: {
      is_pick: {
        description: "Boolean indicating whether the choice is a pick or a ban",
        type: "boolean",
      },
      hero_id: {
        description: "The hero ID",
        type: "integer",
      },
      team: {
        description: "The team that picked or banned the hero",
        type: "integer",
      },
      order: {
        description: "The order of the pick or ban",
        type: "integer",
      },
    },
  },
},
```

There may be many instances such as this. My goal is to be able to use the openapi generator, so I'll include all changes until the generator succeeds. Maybe it's a couple hours work, or maybe I'll only get around to it next weekend. But in the meantime, everything else looks ready for review. @howardchung

File: test/test.js [hunk 1]
```diff
@@ -161,15 +161,29 @@ before(function setup(done) {
         async.series(
           [
             (cb) => {
-              es.indices.delete(
+              es.indices.exists(
```
ff137, line 164: Purpose of this change is because .delete throws a ResponseException in the test if the index does not yet exist. So now only call delete if it exists. Either this, or the delete should be moved to the end, after index has been created.

ff137:
The `openapi-generator-cli` succeeds when the openapi version is set to `3.0.3`. Throws error with `3.1.0`, and seems to be an issue with the generator, not the spec.

So I'm happy to say this upgrade is ready for review. I recommend a version bump, considering there may be breaking changes

howardchung (approved):

howardchung:
@ff137  it looks like the paths for the docs are now incorrect and omit the /api in the path. Is this easily fixed?

ff137:
@howardchung I don't know. In the OpenApi 3 spec the url is correctly defined:
```
  servers: [
    {
      url: "https://api.opendota.com/api",
    },
```

That includes /api in the path. 

Just went to https://docs.opendota.com/ to check it out, and yeah, first of all there's the warnings:
```
No "type" specified at "#/components/parameters/matchIdParam". Automatically detected: "object"
No "type" specified at "#/components/parameters/accountIdParam". Automatically detected: "object"
No "type" specified at "#/components/parameters/fieldParam". Automatically detected: "object"
No "type" specified at "/paths/~1search/get/parameters/0". Automatically detected: "object"
No "type" specified at "/paths/~1rankings/get/parameters/0". Automatically detected: "object"
No "type" specified at "/paths/~1benchmarks/get/parameters/0". Automatically detected: "object"
No "type" specified at "/paths/~1request~1{jobId}/get/parameters/0". Automatically detected: "object"
No "type" specified at "#/components/parameters/heroIdPathParam". Automatically detected: "object"
No "type" specified at "#/components/parameters/leagueIdPathParam". Automatically detected: "object"
No "type" specified at "#/components/parameters/teamIdPathParam". Automatically detected: "object"
No "type" specified at "/paths/~1records~1{field}/get/parameters/0". Automatically detected: "object"
No "type" specified at "/paths/~1constants~1{resource}/get/parameters/0". Automatically detected: "object"
```

So that'll need to be patched.

When I clicked on `GET /heroes` as an example, it indeed takes me to https://api.opendota.com/heroes instead of https://api.opendota.com/api/heroes.

Seems to me like an issue with the documentation rendering tool, because the url is right in the spec. It's maybe using an outdated version? Because Swagger 2.0 and OpenApi 3.0 do have different ways of defining the server. But I don't have any knowledge of the doc rendering. What do you think?

ff137:
@howardchung with regards to those warnings, I can confirm that, e.g. for `matchIdParam`, the spec is correct:
```
  matchIdParam: {
    name: "match_id",
    in: "path",
    required: true,
    schema: {
      type: "integer",
    },
  },
```

The type is correctly defined; in OpenApi 3 it must be inside a schema field. 

This further confirms I think the documenting tool is out of date (still expecting Swagger spec)

ff137:
@howardchung You're lucky I'm a machine. Found the prob and fixed it in https://github.com/odota/docs/pull/19

PR #2662, opened 2023-06-21
Add `minRank` and `maxRank` query parameters to `/publicMatches`

ff137:
Adds `minRank` and `maxRank` query parameters and uses `avg_rank_tier` field in public_matches to filter results.

```js
    "/publicMatches": {
      get: {
...
        parameters: [
          { $ref: "#/components/parameters/lessThanMatchIdParam" },
          { $ref: "#/components/parameters/minRankParam" },
          { $ref: "#/components/parameters/maxRankParam" },
          { $ref: "#/components/parameters/mmrAscendingParam" },
          { $ref: "#/components/parameters/mmrDescendingParam" },
        ],
...
        route: () => "/publicMatches",
        func: async (req, res, cb) => {
...
          if (req.query.mmr_ascending) {
            order = "ORDER BY avg_rank_tier ASC NULLS LAST";
          } else if (req.query.mmr_descending) {
            order = "ORDER BY avg_rank_tier DESC NULLS LAST";
          } else {
            order = "ORDER BY match_id DESC";
            moreThan = 0;
          }
          let minRank= req.query.min_rank
            ? `AND avg_rank_tier >= ${req.query.min_rank}`
            : "";
          let maxRank= req.query.max_rank
            ? `AND avg_rank_tier <= ${req.query.max_rank}`
            : "";

          db.raw(
            `
          WITH match_ids AS (SELECT match_id FROM public_matches
          WHERE TRUE
          AND match_id > ?
          AND match_id < ?
          ${minRank}
          ${maxRank}
          ${order}
          LIMIT 100)
          SELECT * FROM
...
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "opendota-core",
-  "version": "21.0.0",
+  "version": "21.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "opendota-core",
-      "version": "21.0.0",
+      "version": "21.0.1",
       "license": "MIT",
       "dependencies": {
         "@elastic/elasticsearch": "^7.17.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "opendota-core",
   "description": "Open source Dota data platform",
-  "version": "21.0.0",
+  "version": "21.0.1",
   "license": "MIT",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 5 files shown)


builder-247 (commented):
I'm pretty sure the mmr field isn't being updated anymore, medal should be used instead

ff137:
> I'm pretty sure the mmr field isn't being updated anymore, `average_rank` should be used instead

Oh, ok. I was just going off the response from: https://api.opendota.com/api/publicMatches

In that case, avg_mmr should probably be removed from the response? To avoid confusion. Or, avg_mmr should be implemented again? Is there a reason it's not being updated anymore?

ff137:
@howardchung I notice the indenting in the spec description causes it to render like this:
image

Included a fix for that

howardchung (approved):

Issue #2663, opened 2023-06-22
itemPopularity endpoint fails for some heroes


howardchung:
e.g. https://api.opendota.com/api/heroes/5/itemPopularity

howardchung:
Checked the logs and this parseInt is crashing:
```
/usr/src/store/queries.js:318
        const time = parseInt(item.time, 10);
                                   ^

TypeError: Cannot read properties of null (reading 'time')
    at /usr/src/store/queries.js:318:36
    at Array.map (<anonymous>)
    at /usr/src/store/queries.js:317:8
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
    ```

@ff137

ff137:
#2664

PR #2664, opened 2023-06-23
Handle null item times and adjust cost thresholds in `getHeroItemPopularity`

ff137:
`getHeroItemPopularity` can error if item.time is null. #2663

Also adjust cost thresholds used in defining early, mid, and late game items. See review comments

### DIFF
--- a/routes/responses/schemas/match/PublicMatchesResponse.js
+++ b/routes/responses/schemas/match/PublicMatchesResponse.js
@@ -13,6 +13,27 @@ module.exports = {
       radiant_win: commonProperties.radiant_win,
       start_time: commonProperties.start_time,
       duration: commonProperties.duration,
+      avg_mmr: {
+        type: "integer",
+      },
+      num_mmr: {
+        type: "integer",
+      },
+      lobby_type: {
+        type: "integer",
+      },
+      game_mode: {
+        type: "integer",
+      },
+      avg_rank_tier: {
+        type: "integer",
+      },
+      num_rank_tier: {
+        type: "integer",
+      },
+      cluster: {
+        type: "integer",
+      },
       radiant_team: {
         description: "radiant_team",
         type: "string",
--- a/store/queries.js
+++ b/store/queries.js
@@ -330,7 +330,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
     const midGameItems = countItemPopularity(
       items.filter(
         (item) =>
-          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost > 2000
+          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost >= 1200
       )
     );
     const lateGameItems = countItemPopularity(
--- a/store/queries.js
+++ b/store/queries.js
@@ -324,7 +324,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
     );
     const earlyGameItems = countItemPopularity(
       items.filter(
-        (item) => item.time > 0 && item.time < 60 * 10 && item.cost >= 600
+        (item) => item.time > 0 && item.time < 60 * 10 && item.cost >= 500
       )
     );
     const midGameItems = countItemPopularity(
--- a/store/queries.js
+++ b/store/queries.js
@@ -324,7 +324,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
     );
     const earlyGameItems = countItemPopularity(
       items.filter(
-        (item) => item.time > 0 && item.time < 60 * 10 && item.cost > 700
+        (item) => item.time > 0 && item.time < 60 * 10 && item.cost >= 600
       )
     );
     const midGameItems = countItemPopularity(
--- a/store/queries.js
+++ b/store/queries.js
@@ -312,7 +312,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
 
     const items = purchaseLogs.rows
       .flatMap((purchaseLog) => purchaseLog.purchase_log)
-      .filter((item) => item.key !== null)
+      .filter((item) => item !== null && item.key !== null)
       .map((item) => {
         const time = item.time ? parseInt(item.time, 10) : 0;
         const { cost, id } = constants.items[item.key];
--- a/store/queries.js
+++ b/store/queries.js
@@ -315,7 +315,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
       .flatMap((purchaseLog) => purchaseLog.purchase_log)
 
       .map((item) => {
-        const time = parseInt(item.time, 10);
+        const time = item.time ? parseInt(item.time, 10) : 0;
         const { cost, id } = constants.items[item.key];
         return { cost, id, time };
       });
--- a/store/queries.js
+++ b/store/queries.js
@@ -334,7 +334,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
       )
     );
     const lateGameItems = countItemPopularity(
-      items.filter((item) => item.time >= 60 * 25 && item.cost > 4000)
+      items.filter((item) => item.time >= 60 * 25 && item.cost >= 2000)
     );
 
     return cb(null, {
--- a/store/queries.js
+++ b/store/queries.js
@@ -321,7 +321,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
       });
 
     const startGameItems = countItemPopularity(
-      items.filter((item) => item.time <= 0)
+      items.filter((item) => item.time <= 0 && item.cost <= 600)
     );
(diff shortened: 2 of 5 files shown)


ff137:
Here is the existing logic:
```js
    const startGameItems = countItemPopularity(
      items.filter((item) => item.time <= 0)
    );
    const earlyGameItems = countItemPopularity(
      items.filter(
        (item) => item.time > 0 && item.time < 60 * 10 && item.cost > 700
      )
    );
    const midGameItems = countItemPopularity(
      items.filter(
        (item) =>
          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost > 2000
      )
    );
    const lateGameItems = countItemPopularity(
      items.filter((item) => item.time >= 60 * 25 && item.cost > 4000)
    );
```

My first proposition is that start game items is redefined: 
```js
    const startGameItems = countItemPopularity(
      items.filter((item) => item.time <= 0 && item.cost <= 600)
    );
```
because, apparently `null <= 0` will return true in js. So, in case a null time is recorded for a lategame item, we can add `item.cost <= 600` to filter any of those potential edge cases. 600 is the starting gold.

My other proposition is to adjust the item.cost filters used in mid and late game items. Because, 2000 and 4000 feel like really high thresholds, so will discard a lot of other items. My recommendation would be:
```js
    const midGameItems = countItemPopularity(
      items.filter(
        (item) =>
          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost > 1200
      )
    );
    const lateGameItems = countItemPopularity(
      items.filter((item) => item.time >= 60 * 25 && item.cost > 1800)
    );
```

@howardchung what do you think?

howardchung:
I feel like if we want to check for null, we should explicitly do that check rather than relying on the 600 gold filter (since that could change)

store/queries.js
```diff
@@ -315,7 +315,7 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
       .flatMap((purchaseLog) => purchaseLog.purchase_log)
 
       .map((item) => {
-        const time = parseInt(item.time, 10);
+        const time = item.time ? parseInt(item.time, 10) : 0;
```
howardchung, line 318: does item?.time work?

Also I think the very next line is going to encounter the same crash when we try item.key. So maybe we should be using filter(Boolean) to remove all null items before we try doing something

ff137: `item.time ? parseInt(item.time, 10) : 0` will default to 0 when item.time is null

ff137: I agree filtering any null would make sense. I'm just curious what would cause the null `item.time`. If it only ever happens for pre-game items, then defaulting to 0 would make sense. If it can happen to late game items, then the item.cost <= 600 would filter that. But would also make sense to just remove all nulls from the start, assuming it's not too prevalent

store/queries.js
```diff
@@ -311,31 +311,30 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
...
       });
 
     const startGameItems = countItemPopularity(
-      items.filter((item) => item.time <= 0)
+      items.filter((item) => item.time <= 0 && item.cost <= 600)
     );
     const earlyGameItems = countItemPopularity(
       items.filter(
-        (item) => item.time > 0 && item.time < 60 * 10 && item.cost > 700
+        (item) => item.time > 0 && item.time < 60 * 10 && item.cost >= 600
```
ff137, line 327: >= 600 instead of >700

Essentially will now add: Bottle (675), Voodoo Mask, Ring of Health, Voice Stone (700)

store/queries.js
```diff
@@ -311,31 +311,30 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
...
       )
     );
     const midGameItems = countItemPopularity(
       items.filter(
         (item) =>
-          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost > 2000
+          item.time >= 60 * 10 && item.time < 60 * 25 && item.cost >= 1200
```
ff137, line 333: `>= 1200 instead of `>2000`

Essentially will now add: Cornucopia, Point Booster (1200), Arcane Boots, Talisman of Evasion (1300), Claymore (1350), Aghs Shard, Platemail, Perseverance, Pavise (1400), Ghost Scepter, Oblivion Staff, Phase Boots (1500), Veil of Discord (1525), Mithril Hammer (1600), Drum of Endurance (1650), Vanguard (1700), Mask of Madness, Mekansm (1775), Dragon Lance (1900), Crystalys (1950), Hyperstone (2000).

howardchung:
The LIKE query is going to be slow because there isn't a text index on that column.

store/queries.js
howardchung: if item is null then item.key will crash. I think you want .filter(Boolean) here.

ff137: Ok, I added `.filter((item) => item !== null && item.key !== null)`

ff137: null time will then still default to 0, so if it was a pre-game purchase it'll still be counted by :
```js
    const startGameItems = countItemPopularity(
      items.filter((item) => item.time <= 0 && item.cost <= 600)
    );
```

store/queries.js
ff137: @howardchung last thing I'm wondering about is if there's any justification to reducing this threshold further, maybe to >=500, to include items like boots of speed, bracer, null talisman, wraith band, chainmail.

Reducing to >= 400 would include things like magic wands, gloves of haste, blades of attack, belt of strength, etc.

So it's not a HUGE amount of items being added by reducing threshold, but it's up to you if it makes sense to include magic wand, blades of attack, etc as "early game items". I don't see why not. I think 500 threshold makes most sense to include bracers/etc. Maybe 450 / 400 if you're okay with including wand, band of elvenskin, robe of magi, etc

ff137: Like it may be interesting to see how common boots of speed is in first 10 minutes for a hero. Should be almost every game, but will be differences across heroes. And bracer/wraith/null builds are quite common too, would be interesting for stats to see that too.

store/queries.js
howardchung: I still think .filter(Boolean) is simpler and does what we want there. But if you want to check item and item.key specifically then use != to include checking for undefined

ff137: In the latest changes I added a filter statement: https://github.com/odota/core/pull/2664/files#diff-6483f8be8db993ad2d2cb82be6da1067d8cb8c925a8559e532e063aefa4f5047R315

ff137: We can add filtering to remove cases that have null `item.time` too. I just can't know what makes most sense, given I can't see how many cases this filters, and what causes those cases.

If item.time defaults to 0, then it can potentially only be counted in starting items, and only if it's less than 600 gold. That feels to me like an adequate solution. But I'm open to just remove null cases for time too, it won't change much

ff137:
@howardchung thoughts on these changes?

ff137 (commented):
@howardchung I noticed that PublicMatchesResponse had some missing params in its definition. So, patching that in this PR

routes/responses/schemas/match/PublicMatchesResponse.js
```diff
@@ -13,6 +13,27 @@ module.exports = {
...
+      },
+      avg_rank_tier: {
+        type: "integer",
+      },
+      num_rank_tier: {
+        type: "integer",
+      },
+      cluster: {
+        type: "integer",
+      },
```
ff137, line 36: here*

howardchung:
My only blocking concern is changing the filter condition to:
`.filter((item) => item && item.key)`

ff137:
> My only blocking concern is changing the filter condition to: `.filter((item) => item && item.key)`

How do you mean? I've added: `.filter((item) => item !== null && item.key !== null)`
Do you want to add `item.time !== null`?

howardchung:
That depends on the behavior we want if time is null/undefined. If we want to default it to 0, then we shouldn't add it.

But the change I'm asking for is either to just check for item && item.key or item != null && item.key != null to handle undefined (which may not be possible, but I figure it's safer)

store/queries.js
```diff
@@ -311,31 +311,30 @@ function getHeroItemPopularity(db, redis, heroId, options, cb) {
     }
 
     const items = purchaseLogs.rows
-
       .flatMap((purchaseLog) => purchaseLog.purchase_log)
-
+      .filter((item) => item && item.key && item.time)
```
ff137, line 315: @howardchung I edited as follows

howardchung (approved):

store/queries.js
ff137: makes sense to not need `!== null` check - I'm new to javascript so wasn't sure on diff between the two, but asserting truthiness is probably better than only asserting not null. Filtering by time also makes sense because should be an anomaly

ff137:
@howardchung thanks 👍

PR #2670, opened 2023-09-15
:art: Refactor: move route func's to separate handlers

ff137:


howardchung (approved):

howardchung:
This will need a rebase. But my intent was for queries.ts to hold most of the business logic as well as provide testable interfaces. We don't necessarily need to split each set of functions into a separate file, I think

PR #2679, opened 2023-12-03
Rework cosmetics service

builder-247:
* Changed `items_game.txt` source from SteamDB to d2vpkr, since it's no longer tracked there
* Retrieve CDN URLs from my tool https://github.com/builder-247/node-dota2-cdn, I intend to automate this further
* Rewrite the service to use promises instead of callbacks

### DIFF
--- a/svc/cosmetics.mjs
+++ b/svc/cosmetics.mjs
@@ -1,75 +1,54 @@
-// NOT WORKING: Updates game cosmetic items in the database
+// Updates game cosmetic items in the database
 import vdf from 'simple-vdf';
-import async from 'async';
 import db from '../store/db.mjs';
 import utility from '../util/utility.mjs';
 import queries from '../store/queries.mjs';
-const { invokeInterval, cleanItemSchema } = utility;
-function doCosmetics(cb) {
-  utility.getData(
-    {
-      url: 'https://raw.githubusercontent.com/SteamDatabase/GameTracking-Dota2/master/game/dota/pak01_dir/scripts/items/items_game.txt',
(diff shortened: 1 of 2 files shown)


svc/cosmetics.mjs
```diff
@@ -1,75 +1,54 @@
...
+        item_id: item.item_id,
+      });
     }
-  );
+
+    const funcs = Object.keys(itemData.items_game.items).map(
+      (i) => () => processItem(i)
+    );
+    await eachLimit(funcs, 10);
+  } catch (error) {
```
howardchung, line 49: I think since we're at top level we can just not try catch here and let it crash. The process manager will restart it

util/utility.mjs
```diff
@@ -818,6 +812,23 @@ function invokeInterval(func, delay) {
     });
   })();
 }
+
+/*
+ * Promise replacement for async.eachLimit
+ * Executes funcs at most limit at once
+ */
+async function eachLimit(funcs, limit) {
```
howardchung, line 820: can we make this work an array of promises instead of funcs

howardchung:
Awesome, thanks for fixing this! It's been great working with modernized code so far--debating trying to go to TypeScript at some point haha

util/utility.mjs
howardchung, line 820: you can export this directly as well

builder-247:
> Awesome, thanks for fixing this! It's been great working with modernized code so far--debating trying to go to TypeScript at some point haha

It sure was interesting issue, didn't expect to spend my day in a hex editor when I started on this. Learned a lot nonetheless.

PR #2694, opened 2023-12-22
Hide data for players with match history disabled

### LINKED ISSUE #2645: [Request] Implement Private Profiles
**Problem:** Opendota still has my historical data even if I have `Expose Match Data` setting off. Even though it's *my* fault to enable it one year ago for a moment - it is still a problem that I cannot make *my* data go private - I just don't want my match history/past stats to be shown to everybody. This was also requested many times in #general discord channel.

**Motivation:** With recent Valve [news on cheaters](https://steamcommunity.com/games/dota2/announcements/detail/3677788723152833274) and [draft anonymity](https://steamcommunity.com/games/dota2/announcements/detail/3659774959178253451) I think it's clear that Valve want to support anonymity and respect private data. So it would be nice for OpenDota to follow the example. Maybe there is some law thing too, but I don't want to nerd it out; it just would be humanly-nice to be able to set my opendota to private. 

Valve made profiles hidden during the draft but in theory players still can google my name/see my avatar in Opendota and get the stats. So next step for better anonymity is either for Valve to hide names/avatars in draft or for Opendota to implement Private Profiles.

**Solution:** For example, Stratz, Dotabuff already respect the Expose setting and hide my profile data from others. 
image
image

* Dotabuff approach would be perfect: where the data is hidden from others but I can still access it if I log in.
* Stratz approach to hide the data from everybody including me is also fine. (edit: now you can view your data if you login in stratz too)

OverwolfDotaPlus also respected Valve decision and removed sniping-people data in the draft stage too. 

**Alternative solution:** odota/web#2883 where OP suggests manual opt-in/out, but I think just following Stratz/Dotabuff example would be more intuitive.

howardchung:
fixes #2645 

Return an empty array for player endpoints if the player has unchecked Expose Public Match Data and we've picked up the change via profile refresh.

If the player is logged in and matches the account_id of the player being requested, return data normally.

### DIFF
--- a/routes/api.ts
+++ b/routes/api.ts
@@ -4,6 +4,7 @@ import spec from './spec';
 import { readCache } from '../store/cacheFunctions';
 import { redisCount } from '../util/utility';
 import redis from '../store/redis';
+import db from '../store/db';
 
 //@ts-ignore
 const api: Router = new Router();
@@ -32,7 +33,7 @@ api.use('/players/:account_id/:info?', async (req, res, cb) => {
   }
 });
 // Player endpoints middleware
-api.use('/players/:account_id/:info?', (req, res, cb) => {
+api.use('/players/:account_id/:info?', async (req, res, cb) => {
   if (!Number.isInteger(Number(req.params.account_id))) {
     return res.status(400).json({ error: 'invalid account id' });
   }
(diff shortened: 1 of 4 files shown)


routes/api.ts
```diff
@@ -53,6 +54,9 @@ api.use('/players/:account_id/:info?', (req, res, cb) => {
     filterCols = filterCols.concat(filterDeps[key as FilterType] || []);
   });
   const sortArr = (req.query.sort || []) as (keyof ParsedPlayerMatch)[];
+  const privacy = await db.raw('SELECT fh_unavailable FROM players WHERE account_id = ?', [req.params.account_id]);
```
howardchung, line 57: Need a try/catch here to handle exceptions

PR #2700, opened 2023-12-29
switch to daily rate limit

howardchung:
cc @albertcui 

Since I'm changing the Redis key name this will effectively reset all usage counts for the month.

We should also update docs and UI

### DIFF
--- a/svc/web.ts
+++ b/svc/web.ts
@@ -20,6 +20,7 @@ import db from '../store/db';
 import redis from '../store/redis';
 import config from '../config';
 import {
+  getEndOfDay,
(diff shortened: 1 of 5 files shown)


svc/web.ts
```diff
@@ -181,7 +182,7 @@ app.use((req, res, cb) => {
     });
     if (!res.locals.isAPIRequest) {
       res.set(
-        'X-Rate-Limit-Remaining-Month',
+        'X-Rate-Limit-Remaining-Day',
```
howardchung, line 185: change "monthly api limit exceeded" to "daily"

Issue #2646, opened 2023-03-07
fix items to DB service


maqic:
### Reproduction Method
```bash
curl "https://api.steampowered.com/IEconDOTA2_570/GetGameItems/v1?key=xxx"
```

### Error Detail

```
HTTP/1.1 404 Not Found
Server: nginx
Content-Type: text/html; charset=UTF-8
Content-Length: 149
Expires: Tue, 07 Mar 2023 02:38:26 GMT
Cache-Control: max-age=0, no-cache, no-store
Pragma: no-cache
Date: Tue, 07 Mar 2023 02:38:26 GMT
Connection: close

<html><head><title>Not Found</title></head><body><h1>Not Found</h1>Method 'GetGameItems' has been retired in interface 'IEconDOTA2_570'</body></html>
```

### Effects

https://github.com/odota/core/blob/e1c0de64d6edd2cdbf301f9cad882428e2965e56/util/utility.js#L111

cozenom:
GetGameItems also not working for IEconDOTA2_205790 anymore.

PR #2764, opened 2024-02-25
Use new api endpoint for items

builder-247:
Closes https://github.com/odota/core/issues/2646

### DIFF
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -296,7 +296,6 @@ CREATE TABLE IF NOT EXISTS items(
   secret_shop smallint,
   side_shop smallint,
   recipe smallint,
-  neutral_item_tier smallint,
   localized_name text
 );
 
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -296,6 +296,7 @@ CREATE TABLE IF NOT EXISTS items(
   secret_shop smallint,
   side_shop smallint,
   recipe smallint,
+  neutral_item_tier smallint,
   localized_name text
 );
 
--- a/svc/items.ts
+++ b/svc/items.ts
@@ -1,4 +1,5 @@
 // Updates game items in the database
+import { items } from 'dotaconstants';
 import db from '../store/db';
 import { upsert } from '../store/insert';
 import {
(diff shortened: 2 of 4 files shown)


svc/items.ts
```diff
@@ -8,20 +8,22 @@ import {
...
   const body = await getSteamAPIData(container.url);
-  if (!body || !body.result || !body.result.items) {
+  if (!body || !body.result || !body.result.data) {
     throw new Error('invalid body');
   }
   await Promise.all(
-    body.result.items.map((item: any) =>
+    body.result.data.itemabilities.map((item: any) => {
+      item.localized_name = item.name_english_loc;
+      // NOTE: properties cost, secret_shop, side_shop and recipe are no longer present
```
howardchung, line 21: If we don't have this data anymore is this an adequate replacement for the endpoint?

I think we do have the data from dotaconstants, so maybe that would be a better source?

sql/create_tables.sql
```diff
@@ -296,6 +296,7 @@ CREATE TABLE IF NOT EXISTS items(
   secret_shop smallint,
   side_shop smallint,
   recipe smallint,
+  neutral_item_tier smallint,
```
howardchung, line 299: this is no longer available right?

svc/items.ts
```diff
@@ -8,20 +9,24 @@ import {
...
     language: 'english',
   });
   const body = await getSteamAPIData(container.url);
-  if (!body || !body.result || !body.result.items) {
+  if (!body || !body.result || !body.result.data) {
     throw new Error('invalid body');
   }
   await Promise.all(
-    body.result.items.map((item: any) =>
+    body.result.data.itemabilities.map((item: any) => {
```
howardchung, line 20: is this still under itemabilities?

howardchung: Oh I see, this is merging the two data sources. Maybe we can just use dotaconstants only?

builder-247: dotaconstants doesn't currently have the neutral tier property, I suggest merging this as a temporary fix and I can move it to only use dotaconstants when I have the time

howardchung: The neutral tier data doesn't currently exist right? If we don't need it then I can avoid having to make the schema change

builder-247: Correct, just figured it would be useful. I don't think the items table is even being used internally at all, at least I wasn't able to find any uses.

howardchung: It is not--the only use for it is by users joining it in custom SQL explorer queries. That low usage is why it's been broken for so long.

howardchung (approved):

PR #2803, opened 2024-05-28
Add selected Facets to matches

builder-247:
Since the web API endpoint is broken I haven't been able to fully test this but the retriever seems to get the correct data.


howardchung:
I think there's already a new field "hero_variant" in the API response that has this data. You can check it with the scanner since GetMatchDetails is currently broken

routes/playerFields.ts
```diff
@@ -52,6 +52,7 @@ export const matchesCols = [
   'average_rank',
   'leaver_status',
   'party_size',
+  'hero_variant',
```
howardchung, line 55: Cassandra schema needs to be updated if we want to return it in the player calls

sql/create_tables.cql
```diff
@@ -65,6 +65,7 @@ CREATE TABLE IF NOT EXISTS player_caches (
   item_5 text,
   item_neutral text,
   average_rank int,
+  hero_variant int,
```
howardchung, line 68: This should be text, since it's only one byte it's more efficient to store as text

PR #2808, opened 2024-08-25
feat: add neutral_token drops data

Antosik:
Adds the ability to track the neutral tokens drops
Related odota/parser PR - https://github.com/odota/parser/pull/72

### DIFF
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -137,7 +137,8 @@ CREATE TABLE IF NOT EXISTS player_matches (
   roshans_killed int,
   observers_placed int,
   party_size int,
-  net_worth int
+  net_worth int,
+  neutral_tokens_log json[]
 );
 CREATE INDEX IF NOT EXISTS player_matches_account_id_idx on player_matches(account_id) WHERE account_id IS NOT NULL;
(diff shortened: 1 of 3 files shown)


sql/create_tables.sql
```diff
@@ -137,7 +137,8 @@ CREATE TABLE IF NOT EXISTS player_matches (
   roshans_killed int,
   observers_placed int,
   party_size int,
-  net_worth int
+  net_worth int,
+  neutral_tokens_log json[]
```
howardchung, line 141: cql also needs to be updated

sql/create_tables.cql
```diff
@@ -66,6 +66,7 @@ CREATE TABLE IF NOT EXISTS player_caches (
   item_neutral text,
   average_rank int,
   hero_variant text,
+  neutral_tokens_log text,
```
howardchung, line 69: player_matches not player_caches

howardchung: Oh wait we recently switched to the blob model. So this change is actually not needed in the cql

Antosik:
@howardchung 🙏

howardchung (approved):

PR #2861, opened 2025-03-14
:recycle: Refactor timestamp handling to use UTC

ff137:
Following up from this discord chat: https://discord.com/channels/144629812982448128/144629812982448128/1349940912205070380

Timestamp processing should use UTC, and `moment()` uses local time be default. So this PR just replaces all usages of `moment()` with `moment.utc()`

### DIFF
--- a/fetcher/getGcData.ts
+++ b/fetcher/getGcData.ts
@@ -52,12 +52,12 @@ async function saveGcData(
   redis.hincrby('retrieverSteamIDs', steamid, 1);
   redis.expireat(
     'retrieverSteamIDs',
-    moment().startOf('day').add(1, 'day').format('X'),
+    moment.utc().startOf('day').add(1, 'day').format('X'),
   );
   redis.hincrby('retrieverIPs', ip, 1);
   redis.expireat(
     'retrieverIPs',
-    moment().startOf('day').add(1, 'day').format('X'),
+    moment.utc().startOf('day').add(1, 'day').format('X'),
   );
   if (headers['x-match-noretry']) {
     // Steam is blocking this match for community prediction, so return error to prevent retry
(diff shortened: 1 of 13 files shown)


howardchung:
The production instance already uses UTC so I think this won't affect anything there, but could matter for people running their own copy.

ff137:
Makes sense - then it's just for safety / consistency. 

Btw I looked up that `Date.now()` uses UTC by default. So they're good. Just `moment()` uses local datetime by default

fetcher/getGcData.ts
```diff
@@ -52,12 +52,12 @@ async function saveGcData(
   redis.hincrby('retrieverSteamIDs', steamid, 1);
   redis.expireat(
     'retrieverSteamIDs',
-    moment().startOf('day').add(1, 'day').format('X'),
+    moment.utc().startOf('day').add(1, 'day').format('X'),
```
howardchung, line 55: Actually irrelevant whenever we use format X since unix timestamp has no time zone

svc/parser.ts
```diff
@@ -156,7 +156,7 @@ async function parseProcessor(job: ParseJob, metadata: JobMetadata) {
     const message = c[colors[type]](
       `[${new Date().toISOString()}] [parser] [${type}: ${
         end - start
-      }ms] [api: ${apiTime}ms] [gcdata: ${gcTime}ms] [parse: ${parseTime}ms] [queued: ${moment(
+      }ms] [api: ${apiTime}ms] [gcdata: ${gcTime}ms] [parse: ${parseTime}ms] [queued: ${moment.utc(
```
howardchung, line 159: IDK if we want to do this here since we're creating from timestamp already

fetcher/getGcData.ts
ff137: Out of curiosity, I just wrote a test for this one, and they are indeed different -- because moment() and moment.utc() start with different baseline value -- adding one day and only then calling format('x'), means the results have different values

ff137: With the help of GPT:
```ts
describe('Moment.js UTC vs Local Time', () => {
  it('should show the difference between local and UTC start of day', () => {
    const localStartOfDay = moment().startOf('day').add(1, 'day').format('X');
    const utcStartOfDay = moment.utc().startOf('day').add(1, 'day').format('X');

    // Assert that the two timestamps are different
    assert.notEqual(localStartOfDay, utcStartOfDay, `Local and UTC start of day should differ. Local: ${localStartOfDay}, UTC: ${utcStartOfDay}`);
  });
}
```

svc/parser.ts
ff137: This was also one that I was wondering about, but it seems to make no difference. moment vs moment.utc just impacts the timezone info. .e.g here's the difference I get:
```
moment(timestamp) - Sat Mar 15 2025 09:57:03 GMT+0200, 
moment.utc(timestamp) - Sat Mar 15 2025 07:57:03 GMT+0000
```

`.fromNow()` is called, and then either one ends up the same

ff137: So it's not necessary to add .utc() here, but it doesn't make a difference. It'll just be for consistency to do `.utc()`

PR #2888, opened 2025-05-24
Fix averageMedal calculation

mdiller:
The `averageMedal` method used for computing the average of a set of ranks is very slightly incorrect. We basically treat the ranks as if they are base 5 numbers, and average them based on that. The issue is that rank  are only *almost* base 5 numbers. the second digit is 1-5, when in a base 5 number it would be 0-4. This PR updates the method to account for that so that they get averaged properly.

Here is an example of a scenario where the current implementation is incorrect: 
```js
var ranks = [ 14, 15, 21 ]; // Herald 4, Herald 5, Guardian 1

// EXPECTED AVERAGE: 15 (Herald 5)

// before this PR
console.log(averageMedal(ranks)) // PRINTS: 21 (Guardian 1)

// after this PR
console.log(averageMedal(ranks)) // PRINTS: 15 (Herald 5)
```

### DIFF
--- a/util/utility.ts
+++ b/util/utility.ts
@@ -356,10 +356,10 @@ export function average(data: number[]) {
  * */
 export function averageMedal(values: number[]) {
   const numStars = values.map(
-    (value) => Number(String(value)[0]) * 5 + (value % 10),
+    (value) => Number(String(value)[0]) * 5 + (value % 10) - 1,
   );
-  const avgStars = numStars.reduce((a, b) => a + b, 0) / numStars.length;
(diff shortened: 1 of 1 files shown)


howardchung (approved):

howardchung:
Noticed that this function will return weird values for Immortal players, e.g.

```
    const test2 = [80, 80, 80];
    assert.equal(averageMedal(test2), 75);
 ```
 
This is probably because Immortal doesn't have stars and so doesn't fit the pattern of the other numbers

PR #2955, opened 2026-05-01
Make Postgres usage easier to debug and safer when writes fail

ff137:
**What we’re changing**

OpenDota talks to PostgreSQL from several places at once: the main API, the SQL explorer, and background workers that pull jobs from a queue. This PR does three small things:

1. **Give each kind of connection a clear name** so that when someone looks at database activity, they can tell whether traffic is coming from the normal API, the explorer, or a queue worker—not just a bunch of anonymous connections.

2. **Add optional knobs for timeouts on the main database pool** (off by default). If the team turns them on via environment variables, queries and idle-in-transaction sessions can be capped so a stuck job is less likely to hold a connection forever.

3. **When a batch of database writes fails midway**, we **cancel that batch properly** (rollback) in the places we touched, instead of leaving the database session in an awkward state. That reduces risk of pool weirdness after errors.

**Why we think this helps**

We’ve seen symptoms that look like the database running out of capacity or connections (`too many clients already`, slow periods that improve after restarts). That can come from many processes sharing one server, heavy endpoints, or errors mid-write. This PR doesn’t try to fix that—it **improves observability** (named connections), adds **optional guardrails** (timeouts you can enable gradually), and fixes a **real footgun** (failed transactions not always rolled back). Those are low-risk, reviewable steps before bigger changes (explorer limits, worker concurrency, etc.).

**How it works**

- Each connection type sets a PostgreSQL `application_name` like `odota:<service>:knex`, `...:explorer`, or `...:queue:<queue>:<slot>`.
- New env vars (`POSTGRES_STATEMENT_TIMEOUT_MS`, `POSTGRES_LOCK_TIMEOUT_MS`, `POSTGRES_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_MS`) are **empty by default**; when set to a number, new connections run the matching `SET ...` so ops can tune without code changes.
- `insertMatch`, the rater loop, and Stripe subscriber sync wrap their manual transactions in **try / catch**, call **rollback** on failure, then rethrow so existing behavior and error handling stay the same—only the cleanup path is explicit.

**What we’re not doing here**

- No change to public API behavior, explorer SQL surface, or rate limits.
- No mandatory timeout rollout—defaults stay “off” until you set env vars.

**Suggested review focus**

- Are the `application_name` strings and env var names acceptable for your deployment?
- Any objection to enabling timeouts in staging first (e.g. statement timeout only)?

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/svc/store/db.ts
+++ b/svc/store/db.ts
@@ -50,19 +50,20 @@ function knexPoolSessionStatements(): string[] {
 const knexAppName = pgApplicationName("knex");
 const sessionStmts = knexPoolSessionStatements();
 
+const knexTimeoutSummary = sessionStmts.length
+  ? [
+      `statement=${config.POSTGRES_STATEMENT_TIMEOUT_MS || "0"}ms`,
+      `lock=${config.POSTGRES_LOCK_TIMEOUT_MS || "0"}ms`,
+      `idle_in_tx=${config.POSTGRES_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_MS || "0"}ms`,
+    ].join(" ")
+  : "off";
+
 console.log(
-  "[POSTGRES] knex connecting role=%s max=%s application_name=%s session_timeouts=%s",
+  "[POSTGRES] knex role=%s max=%s application_name=%s session_timeouts=%s",
   config.APP_NAME || "(unset)",
   config.POSTGRES_MAX_CONNECTIONS,
   knexAppName,
-  sessionStmts.length
-    ? {
-        statement_ms: config.POSTGRES_STATEMENT_TIMEOUT_MS || "(off)",
-        lock_ms: config.POSTGRES_LOCK_TIMEOUT_MS || "(off)",
-        idle_in_tx_ms:
-          config.POSTGRES_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_MS || "(off)",
-      }
-    : "(none)",
+  knexTimeoutSummary,
 );
 
 export const db = knex({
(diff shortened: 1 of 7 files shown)


svc/store/db.ts
```diff
@@ -6,25 +6,81 @@ import util from "node:util";
 
 // remember: all values returned from the server are either NULL or a string
 pg.types.setTypeParser(20, (val) => (val === null ? null : parseInt(val, 10)));
+
+/** pg application_name is limited to 63 bytes; keep ASCII-safe for pg_stat_activity. */
+export function pgApplicationName(role: string): string {
+  const app = config.APP_NAME || "unknown";
+  const safeApp = app.replace(/[^\w.-]/g, "_").substring(0, 40);
```
builder-247, line 13: I feel like having this length validation is an overkill (an LLM-ism, I presume)

howardchung:
I think we could accept the try/catch wrapping in a separate PR--I think the current behavior is that the thrown exception causes the process to crash which should also end the transaction, but probably good to roll it back earlier so we don't have to wait for timeout.

The other changes are a bit more involved so I'd like to hold off on those

svc/store/db.ts
ff137: 100%, it is agent-assisted coding. Here it's one of those "solving a problem before it's a problem" things. So it's good, any name will work now, because it normalises the input. But at the same time, it's not _really_ helpful when it wasn't a problem yet. I didn't ask it to solve it, but it's technically good, and once it's produced, we may as well keep it :man_shrugging:

ff137:
Cherry-picked the rollback changes to #2956

PR #2959, opened 2026-05-17
Exponential backoff in runInLoop on failure

ff137:
Previously any error thrown inside a `runInLoop` callback would propagate
all the way out of the worker module, the dynamic import in index.ts
would reject, and the process would sleep 250ms and exit(1) for PM2 to
restart. During a real outage (e.g. Postgres at max_connections), every
runInLoop-based worker (inserter, scanner, scanner2, rater, monitor,
cleanup, ...) crashes and restarts ~4x per second, each restart
reopening its Knex pool against an already-saturated DB and losing
in-flight Promise.all work. The system has no way to slow down.

This change wraps func() in try/catch, logs the error (still loudly,
via console.error, with a consecutive-failure count), and sleeps with
exponential backoff (250ms base, doubling, capped at 30s) plus jitter
before retrying. On success, the failure counter resets.

After `MAX_CONSECUTIVE_FAILURES` (10) in a row, the worker exits so PM2
can do a clean-slate restart (fresh Knex pool, cleared in-memory state)
rather than retrying at the 30s backoff cap forever. With the backoff
curve this is ~2 minutes of sustained failure before giving up — past
any transient DB blip, but short enough that a genuinely-wedged worker
still surfaces via crash + restart instead of silently retrying. Mirrors
the inserter watchdog's existing `process.exit(1)` pattern.

The existing lastRun:<APP_NAME> health beacon is only refreshed on
success, so a worker stuck in the failure path will still trip
HEALTH_TIMEOUT-based alerting. Workers that fail for non-transient
reasons (bad code, missing env) therefore still surface; they just
don't take the DB down with them on the way out.

### DIFF
--- a/svc/store/queue.ts
+++ b/svc/store/queue.ts
@@ -127,9 +127,7 @@ export async function runReliableQueue(
  * not hammered by tight-loop restarts.
  *
  * After MAX_CONSECUTIVE_FAILURES in a row, exits the process so PM2 can do a
- * clean-slate restart (fresh Knex pool, cleared in-memory state). With the
- * default backoff curve this is ~7.5 minutes of sustained failure before
- * giving up, which is well past any transient DB blip.
+ * clean-slate restart (fresh Knex pool, cleared in-memory state).
  *
  * The lastRun health beacon is only refreshed on success, so the existing
  * HEALTH_TIMEOUT alerting still detects a stuck worker before exit.
--- a/svc/store/queue.ts
+++ b/svc/store/queue.ts
@@ -121,17 +121,26 @@ export async function runReliableQueue(
 
 /**
  * Runs an async function on a loop, waiting the delay between each iteration.
+ *
  * On failure, logs and sleeps with exponential backoff + jitter (capped at 30s)
  * before retrying, so a sick dependency (e.g. Postgres at max_connections) is
- * not hammered by tight-loop restarts. The lastRun health beacon is only
- * refreshed on success, so the existing HEALTH_TIMEOUT alerting still detects
- * a stuck worker.
+ * not hammered by tight-loop restarts.
+ *
+ * After MAX_CONSECUTIVE_FAILURES in a row, exits the process so PM2 can do a
+ * clean-slate restart (fresh Knex pool, cleared in-memory state). With the
+ * default backoff curve this is ~7.5 minutes of sustained failure before
+ * giving up, which is well past any transient DB blip.
+ *
+ * The lastRun health beacon is only refreshed on success, so the existing
+ * HEALTH_TIMEOUT alerting still detects a stuck worker before exit.
+ *
  * @param func
  * @param delay
  */
 export async function runInLoop(func: () => Promise<void>, delay: number) {
   const BASE_BACKOFF_MS = 250;
   const MAX_BACKOFF_MS = 30_000;
+  const MAX_CONSECUTIVE_FAILURES = 10;
   let consecutiveFailures = 0;
   while (true) {
     console.log("running %s", func.name);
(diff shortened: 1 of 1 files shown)


svc/store/queue.ts
```diff
@@ -120,15 +120,40 @@ export async function runReliableQueue(
...
+ * before retrying, so a sick dependency (e.g. Postgres at max_connections) is
+ * not hammered by tight-loop restarts. The lastRun health beacon is only
+ * refreshed on success, so the existing HEALTH_TIMEOUT alerting still detects
+ * a stuck worker.
  * @param func
  * @param delay
  */
 export async function runInLoop(func: () => Promise<void>, delay: number) {
+  const BASE_BACKOFF_MS = 250;
+  const MAX_BACKOFF_MS = 30_000;
```
howardchung, line 140: why is this 30_000?

ff137: That'll cap the backoff duration to 30 seconds

howardchung: Oh I've just never seen the _ separator before but looks like it's just a visual feature

svc/store/queue.ts
```diff
@@ -120,15 +120,40 @@ export async function runReliableQueue(
...
+        func.name,
+        consecutiveFailures,
+        e,
+      );
+      // 2 ** capped so the exponent doesn't overflow on long outages
+      const exp = Math.min(consecutiveFailures, 10);
+      const backoff = Math.min(MAX_BACKOFF_MS, BASE_BACKOFF_MS * 2 ** exp);
+      const jitter = Math.random() * 250;
+      await new Promise((resolve) => setTimeout(resolve, backoff + jitter));
+      continue;
```
howardchung, line 169: Don't we want to restart the process at some point (maybe if we have hit the max failure time?)

svc/store/queue.ts
ff137: Ah yes, it's a TypeScript feature to add separators for long numbers. Compiles to js as a regular number

svc/store/queue.ts
ff137: Good point - the while true loop means it can keep retrying without restart.

I'll add a "max consecutive failures" config

ff137: Looks like my agent prompt decided to commit and push the change itself :joy:

It sets `MAX_CONSECUTIVE_FAILURES` to 20, which means it'll retry for ~7.5 minutes before exiting. I would've gone for 10, bringing it closer to 2--2.5 minutes of failures before exiting.

Edit: I've changed it to `MAX_CONSECUTIVE_FAILURES = 10`.

svc/store/queue.ts
```diff
@@ -146,6 +155,14 @@ export async function runInLoop(func: () => Promise<void>, delay: number) {
...
+      if (consecutiveFailures >= MAX_CONSECUTIVE_FAILURES) {
+        console.error(
+          "%s exceeded %d consecutive failures, exiting for PM2 restart",
+          func.name,
+          MAX_CONSECUTIVE_FAILURES,
+        );
+        process.exit(1);
+      }
       // 2 ** capped so the exponent doesn't overflow on long outages
       const exp = Math.min(consecutiveFailures, 10);
```
howardchung, line 165: should we just reuse the same config value here

ff137: I thought so too just now, but it looks like the goal with `const exp = Math.min(consecutiveFailures, 10);` is to prevent the number from overflowing if `MAX_CONSECUTIVE_FAILURES` was set to some higher number. It's just a bit safer this way.

I was going to drop `exp` and just do `2 ** consecutiveFailures` directly, but I see then it can blow up

ff137: @howardchung any changes you'd like to see?

howardchung (approved):

Issue #2969, opened 2026-07-30
Proposal: benchmarks and hero pick/win stats segmented by rank bracket (and position)


geracosta:
### Problem

`/benchmarks` percentiles are global across all skill levels, but "good GPM on this hero" differs a lot between Herald and Immortal, so the percentile a player gets back is hard to act on. Similarly, `/heroStats` already splits picks/wins by rank bracket, but there is no position dimension (the same hero can be a mid or a support with very different winrates).

### How it works today (from the code)

- `svc/counts.ts` increments Redis keys `{heroId}:{rank}:pick|win:{day}` per scanned match (`rank = floor(avg_rank_tier / 10)`, TTL 8 days), and `/heroStats` sums them.
- `updateBenchmarks` samples a percentage of significant matches into global zsets `benchmarks:{epoch}:{metric}:{heroId}`, with 7 metrics defined in `benchmarksUtil.ts`. `/benchmarks` reads percentiles from there.

### Proposal (incremental, each step works standalone)

1. **Benchmarks by bracket group.** Add a bracket bucket to the zset key: `benchmarks:{epoch}:{metric}:{heroId}:{bracketGroup}`. To contain memory, use 4 groups of two ranks each (Herald-Guardian, Crusader-Archon, Legend-Ancient, Divine-Immortal) instead of 8 individual brackets, so the zset count goes x4 rather than x8. Expose via a `bracket` query param on `/benchmarks`, keeping the current global response as the default.
2. **Position dimension on picks/wins.** The counts job runs on Steam API data (no replay), but per-team farm ranking (GPM/LH) is enough for a rough 1-5 position estimate (related: #1590, which proposes the replay-based version of the same idea). Keys become `{heroId}:{rank}:{pos}:pick|win:{day}` - x5 on keys that already expire after 8 days.
3. **More benchmark metrics.** kills/deaths/assists per min, camps stacked, stuns (already stubbed out in `benchmarksUtil.ts`), support gold via `gold_reasons`. Each new metric is one more zset per hero/bracket.

### Why

I'm building a coaching tool on top of the OpenDota API. "Your GPM vs players of your own bracket on this hero" is far more actionable than a global percentile, and today this kind of segmentation is only available via third-party APIs (Stratz), even though the counts pipeline here already tracks brackets.

Since this has a real memory cost in Redis I wanted to check the design before writing any code. Happy to start with just (1) as its own PR if the cost looks acceptable, and gauge from there.

howardchung:
[...]

PR #2972, opened 2026-08-01
Add rank bracket segmentation to benchmarks

geracosta:
Second step of #2969, following what we discussed there: 8 individual brackets, bucketed by the match's `avg_rank_tier` the same way the hero pick/win counts do it. Matches without `avg_rank_tier` only feed the global sets, and the default `/benchmarks` response is unchanged.

Write side: `updateBenchmarks` zadds the same sampled value into `benchmarks:{epoch}:{metric}:{heroId}:{rank}` next to the existing global key, with the same expiry cycle. The bracket reuses the trailing key slot (`""` / `"turbo"` / `"1"`-`"8"`), so turbo stays global-only and there's no collision.

Read side: `GET /benchmarks?hero_id=X&bracket=5` reads percentiles from the bracketed sets (validated 1-8, 400 otherwise). Omitting the param behaves exactly as today.

One thing to be aware of: since benchmarks live only in Redis, the bracketed sets start filling on deploy — bracketed queries will be sparse until the current epoch completes. Nothing to migrate.

### DIFF
--- a/svc/counts.ts
+++ b/svc/counts.ts
@@ -211,6 +211,8 @@ runReliableQueue("counts", 2, async (job: CountsJob, metadata) => {
 
     async function updateBenchmarks() {
       const turbo = isTurbo(match);
+      // 1 to 8 based on the average rank of the match, as in updateHeroCounts
+      const rank = !turbo && avg ? Math.floor(avg / 10) : null;
       if (
         match.match_id % 100 < Number(config.BENCHMARKS_SAMPLE_PERCENT) &&
         (isSignificant(match) || turbo)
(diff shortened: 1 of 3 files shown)


svc/counts.ts
```diff
@@ -242,6 +244,20 @@ runReliableQueue("counts", 2, async (job: CountsJob, metadata) => {
...
+                if (rank) {
+                  const rankKey = [
+                    "benchmarks",
+                    getStartOfBlockMinutes(
+                      Number(config.BENCHMARK_RETENTION_MINUTES),
+                      0,
+                    ),
+                    key,
+                    p.hero_id,
+                    rank,
```
howardchung: Considering that turbo and rank are different dimensions I think it should not reuse the same position in the key. Maybe you can reuse rKey but just add on the bracket?

geracosta: Done in ff39364 — the bracketed sets now keep the empty global/turbo slot and append the bracket as an extra position: `benchmarks:<block>:<metric>:<hero>::<bracket>`. On the write side it's literally `rkey + ":" + rank`, and the read side builds the same suffix.

howardchung (approved):

PR #2976, opened 2026-08-10
Resolve the first blood victim to a player slot in the objectives

### LINKED ISSUE #1488: Parsed Match is missing who got killed for first blood event
Related to odota/ui#1042,  This match didn't seem to have a "killed" hero for the first blood event.  The first blood should have been the Drow (player_slot 2 IIRC), but nothing was returned.

Replay linked for analysis.

https://github.com/gu3st/replays/blob/master/3280538799.dem

geracosta:
Closes #1488.

The data has been there all along: since 2017 the parser's `handleFirstblood` stores the victim's 0-9 player index in the `CHAT_MESSAGE_FIRSTBLOOD` objective's `key` ([Parse.java emits `player2`, CreateParsedDataBlob keys it](https://github.com/odota/parser/blob/master/src/main/java/opendota/CreateParsedDataBlob.java)), but nothing downstream interprets it — consumers see `"key": "8"` with no way to tell it's the victim.

This annotates the objective at query time in `buildMatch` (same pattern as `position_est`): a `victim_player_slot` field resolved through the players array. Because it's query time, it works retroactively for every already-parsed match — no reparse needed.

Example output for match 8754463676:

```json
{"time": 58, "type": "CHAT_MESSAGE_FIRSTBLOOD", "key": "8", "slot": 0, "player_slot": 0, "victim_player_slot": 131}
```

Validated against two production matches:
- 8754463676 → `victim_player_slot: 131` (Keeper of the Light; the killer's `kills_log` has `npc_dota_hero_keeper_of_the_light` at the same second)
- 8813802531 → `victim_player_slot: 1` (Beastmaster; Lone Druid's `kills_log` shows the kill at 16s and Beastmaster's `killed_by` includes him)

Old parses without the field (`key: "null"`) and non-standard player arrays are guarded — the field is simply absent. Typecheck clean except the two pre-existing Stripe errors on master.

### DIFF
--- a/svc/util/buildMatch.ts
+++ b/svc/util/buildMatch.ts
@@ -1,6 +1,10 @@
 import { heroes } from "dotaconstants";
 import config from "../../config.ts";
-import { computeMatchData, estimatePositions } from "./compute.ts";
+import {
+  annotateFirstbloodVictim,
+  computeMatchData,
+  estimatePositions,
+} from "./compute.ts";
 import { buildReplayUrl, isTurbo } from "./utility.ts";
 import redis, { redisCount } from "../store/redis.ts";
 import db from "../store/db.ts";
@@ -267,6 +271,7 @@ export async function buildMatch(
   };
   computeMatchData(matchResult as ParsedPlayerMatch);
   estimatePositions(matchResult.players as ParsedPlayerMatch[]);
+  annotateFirstbloodVictim(matchResult);
 
   // Save in cache
   if (matchResult && config.ENABLE_MATCH_CACHE) {
--- a/svc/util/compute.ts
+++ b/svc/util/compute.ts
@@ -336,6 +336,27 @@ export function estimatePositions(players: ParsedPlayerMatch[]) {
 }
 
 /**
+ * The parser stores the first blood victim as the victim's 0-9 player index
+ * in the CHAT_MESSAGE_FIRSTBLOOD objective's key. Resolve it to a
+ * victim_player_slot field alongside the killer's player_slot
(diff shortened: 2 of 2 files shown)


howardchung:
Change looks fine.

I'm not entirely sure this fixes the issue described in the original post though. That seems to suggest the replay doesn't have the data in the combat log event for that particular game?

geracosta:
Went back to the original replay (gu3st still hosts it) and parsed it with the current parser. The chat event does have the data: CHAT_MESSAGE_FIRSTBLOOD comes out as {time: 120, slot: 9, key: 2}, and players[2] is the Drow at player_slot 2, which is what the OP said was missing. What the issue is really showing is the combat log side: the killing blow came from a lane creep (the Drow's death at 120 reads killed by npc_dota_creep_badguys_melee, seen via the deaths_log branch), so there's no hero kill event at all. The victim can only come from the chat event key, which is what this PR reads. The match is unparsed on the site right now so it returns nothing today either way; after a reparse this would output victim_player_slot: 2 for it.

howardchung:
This only affects the first blood event right? Seems like the scope of the fix is limited (unless we specifically need to do something with this data)

geracosta:
Yes, first blood only. The site itself has a use for it though: the match Timeline shows a first blood marker and currently infers the victim from the killer's first kills_log entry, which is exactly what breaks in the issue's replay (a creep landed the blow, so there's no kills_log entry and the marker shows nothing). With victim_player_slot it can just show the right hero, I can do that web PR next. And it closes #1488 either way.

howardchung (approved):

PR #2982, opened 2026-09-03
Cover match_id in the player_matches hero index

geracosta:
Hero lookups on player_matches use an index on hero_id alone, so postgres visits the heap for every matching row just to read match_id. On a four hero combination query that is 175k heap visits and the plan shows it is most of the cost:

```
GroupAggregate  (cost=35179.89..36556.85 rows=4837 width=12)
  Group Key: match_id
  Filter: (count(DISTINCT hero_id) = 4)
  ->  Sort  (cost=35179.89..35618.72 rows=175533 width=12)
        Sort Key: match_id
        ->  Index Scan using player_matches_hero_id_idx on player_matches  (cost=0.43..18559.85 rows=175533 width=12)
              Index Cond: (hero_id = ANY ('{86,19,106,100}'::integer[]))
```

Adding match_id turns that into an index only scan. hero_id stays the leading column so anything filtering on it alone keeps working, which makes the old index redundant, and the migration swaps it.

This came from a user hitting explorer timeouts on hero matchup queries on Discord. It is the same shape the Combos page uses (odota/web#3380 rewrites that one), and the index helps any query of this shape, including the ones users write themselves.

I cannot measure the improvement without creating the index, so this is plan reasoning rather than a before/after. player_matches is small, about 2.5M rows for 251k matches, so the extra index should be cheap.


sql/create_tables.sql
```diff
@@ -141,7 +141,7 @@ CREATE TABLE IF NOT EXISTS player_matches (
   neutral_item_history json[]
 );
 CREATE INDEX IF NOT EXISTS player_matches_account_id_idx on player_matches(account_id) WHERE account_id IS NOT NULL;
-CREATE INDEX IF NOT EXISTS player_matches_hero_id_idx on player_matches(hero_id);
+CREATE INDEX IF NOT EXISTS player_matches_hero_id_idx on player_matches(hero_id, match_id);
```
howardchung: index name should be probably be updated to match

geracosta: done in 11d49573. it creates `player_matches_hero_id_match_id_idx` now, which matches how the other multi-column ones are named (`hero_ranking_hero_id_score_idx`, `insert_queue_processed_match_seq_num_idx`).

i also dropped the `ALTER INDEX ... RENAME TO player_matches_hero_id_idx` from the end of the migration, since it was putting the old name back and a fresh install would have ended up with a different name than a migrated one. the migration now creates it under the new name and drops the old index, so both paths land on the same thing.

howardchung (approved):

PR #2983, opened 2026-09-11
Support per-minute derived columns in player matches

Xvie:
`kills_per_min`, `hero_damage_per_min` and `tower_damage_per_min` aren't columns in `player_caches`, so asking for one silently does nothing — `sort=hero_damage_per_min` on `/players/{account_id}/matches` comes back without the field and in arbitrary order, since the projection is filtered down to real Cassandra columns and the in-JS sort then compares `undefined` against `undefined`.

Adds a small `derivedCols` map of dependencies plus a compute function, following the shape of `filterDeps` and reusing the `(value / duration) * 60` form already in `benchmarksUtil.ts`. `getPlayerMatchesWithMetadata` expands a requested derived column into the stored columns it needs and computes it after the merge, which covers `project` and `sort` in one place with no route handler changes. Another column is one more entry in the map.

Two details worth flagging: the dependency-expanded list is what goes to `mergeMatches`, otherwise `pick()` strips the source columns off archived matches and the values come back null for archived players. And the value is null rather than 0 when duration or the source column is missing, so unparsed matches sort to the bottom instead of the top.

**On the approach — happy to switch if you'd rather.** The established pattern here is to compute at write time and store the column, the way `actions_per_min` does, and `compute.ts` already computes `kills_per_min` on every match and discards it for want of a column. That version is a much smaller diff: three `ALTER TABLE`s, two more blocks in `compute.ts`, and none of this file changes. I went with read-time derivation because stored columns would only populate matches inserted after deploy, the way `average_rank` is null on older rows, and a records page is specifically about all-time bests — it would rank only recent matches until a backfill happened. If you'd prefer the columns and are fine with that, or have a backfill path I'm not seeing, I'll rewrite it. A secondary point in favour of deriving: `kills_per_min` keeps the `player_temp` fast path this way, since `kills` and `duration` are both in `cacheableCols`.

`npm run test` passes (87). A `DERIVED COLUMNS` suite covers the module directly — each column's value, that a null or zero `duration` yields null rather than `Infinity`, that a zero source column stays 0 while a missing one is null, that dropping any declared dependency makes the value uncomputable, and that `withDerivedDeps` / `withDerivedCols` pass their input straight through when nothing is requested and don't mutate it when something is. The integration test through `getPlayerMatches` fails against the unmodified `buildPlayer.ts`. The fixture only has one match, so I checked ordering separately against a local instance with a second row added: a 10 min / 9000 hero damage game sorts above a 37 min / 12234 one, and `tower_damage_per_min` reverses that pair.

Measured against a local instance with a synthetic 20,000 match account, master vs this branch, `ENABLE_PLAYER_CACHE` on. A plain `?limit=20` is 27.6 ms on master and 28.1 ms here, so requests that don't ask for a derived column are unaffected. `?sort=hero_damage_per_min` is 206.7 ms against master's 207.7 ms for the equivalent `?sort=hero_damage` — the cost there is the existing fallback off the `player_temp` path for a non-cacheable column, not the derivation. `?sort=kills_per_min` stays on the fast path at 36.2 ms, since `kills` and `duration` are both in `cacheableCols`.

UI side: odota/web#3383


howardchung:
I kind of like maintaining the purity of the projection to only support columns that actually exist in Cassandra. Otherwise there's no clear documentation on exactly what's supported. This is a pretty straightforward computation to do on the client side I think

Xvie:
Makes sense — keeping `project` to columns that actually exist is the clearer contract, and it does work out fine on the client.

Moved it to odota/web#3383. It fetches the history with only the columns the records table renders plus the one the value is computed from, which is 10 fields per row rather than the 18 the default projection returns, then ranks it in a transform alongside the existing `transformTrends` / `transformHistograms`. Comes out around 22 KB gzipped for a 975 match account and 64 KB at 3k, so roughly 21-23 bytes per match.

Closing this. Thanks for taking a look.

Issue #2980, opened 2026-08-21
Segment benchmarks by core/support


geracosta:
Follow-up to the position dimension left open in #2969 (point 2), now that position_est (#2974) and the brackets (#2972) are both in.

Problem: benchmarks are written from api data in counts.ts, and the position estimator needs parsed data (gold_t, lane_role), so it can't run there.

So I measured what the api data alone can do, on the same 100 pro matches from the #1590 study: ranking each team by final gold_per_min gets the exact 1-5 position right only 58% of the time, but it separates cores from supports on 99.4% of players (924/930).

So the proposal: segment the bracket benchmarks by core/support instead of 1-5. That's the split that actually changes the numbers per hero (a support Mirana and a core Mirana shouldn't share a distribution), it works at full volume with no parse dependency, and it only doubles the bracket keys instead of x5. Exact 1-5 could come later from parsed matches only, if that sample turns out dense enough.

Happy to put up the PR if this sounds reasonable.

howardchung:
Sure, that proposal sounds fine.

Issue #1590, opened 2018-01-06
Algorithm to estimate the player's "position" on the team.


joshg2k:
A lot of people refer to the role they're going to play in the game as position 1 through position 5. This would be an extremely valuable data point for my use case, and I imagine it would suit a lot of other developers' needs.

For reference: http://www.liquiddota.com/forum/dota-2-strategy/454943-dota-1-to-5-system

howardchung:
We do something similar in the UI for the analysis feature, although theoretically we could also assign a 1-5 value in the core and return it in the API.  Perhaps we can just order the players by gold farmed?

eclar:
I think that could be an idea, but in this case, you may encounter some data where your offlaner will have mode gold than your mid, or a support more than your offlaner (think about Gh, or yapzor farm). Then, do you consider your qop mid to be a 3 or a 2 if she has less farm than a brew for instance ? Maybe, you could combine it with the position data in the 10 first minutes ? A roamer can be identified as a 4, etc ?

kamikazechaser:
This [info](https://www.reddit.com/r/DotA2/comments/6vdd65/the_positions_15_and_their_roles/) might be relevant too.

geracosta:
I'd like to pick this up. I'm building a coaching tool on top of the OpenDota API and a 1-5 position estimate is one of the few things I still need a third-party API for (Stratz exposes `position`, which is also just a replay heuristic, not ground truth).

Concrete proposal, for parsed matches, computed in `compute.ts` next to the existing `lane_role` / `is_roaming`:

1. Within each team, rank the 5 players by early farm priority: `gold_t`/`lh_t` in the min 10-12 window, plus support indicators (ward purchases from `purchase_log`, support-gold share from `gold_reasons`). Using the early window avoids the problem mentioned above of supports that out-farm cores later in the game.
2. Bottom two of the ranking = supports, top three = cores.
3. Map to numbers using the `lane_role` we already compute: safe-lane core -> 1, mid core -> 2, off-lane core -> 3, off-lane support -> 4 (or roamer via `is_roaming`), safe-lane support -> 5. The farm ranking breaks the tie when two players share a lane.

Naming suggestion: `position_est`, to make explicit it's an estimate.

This would also effectively cover #1844, since core/support is just 1-3 vs 4-5.

If this sounds reasonable I'll put together a PR.

howardchung:
[...]

PR #2984, opened 2026-09-16
Segment benchmarks by core/support

geracosta:
Implements #2980, with the core/support split you okayed there. Cores and supports get their own benchmark sets within each bracket, since a support Mirana and a core Mirana shouldn't share a distribution.

Write side: `updateBenchmarks` ranks each team by `gold_per_min` and calls the top 3 cores and the rest supports, then zadds the sampled value into the role sets next to the existing bracket keys. That's the split I measured in #2980: exact 1-5 from api data alone is only 58%, but core vs support is 99.4% (924/930 players over the same 100 pro matches from the #1590 study). Only computed where there's a bracket, so turbo and matches without `avg_rank_tier` are untouched, and teams that don't have 5 players are skipped.

Following the key layout from ff39364e, the role is its own slot rather than reusing an existing one: `benchmarks:{epoch}:{metric}:{heroId}:{turbo}:{bracket}:{role}`. Existing keys come out byte-identical, so this only doubles the bracket sets and nothing collides.

Read side: `GET /benchmarks?hero_id=X&bracket=5&role=core`. Validated against core/support, and 400 if role comes without a bracket, since the role only segments the bracketed sets. Omitting both behaves exactly as today.

Same caveat as #2972: the sets start filling on deploy, so role queries will be sparse until the current epoch completes. Nothing to migrate.


svc/util/queries.ts
```diff
@@ -108,11 +108,16 @@ export async function getHeroItemPopularity(heroId: string) {
...
+  bracket?: number | null,
+  role?: string | null,
+) {
   const ret: AnyDict = {};
   const arr = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.95, 0.99];
-  // Bracketed sets add the 1-8 rank bracket after the global/turbo slot
-  const suffix = bracket ? `:${bracket}` : "";
+  // Bracketed sets add the 1-8 rank bracket after the global/turbo slot, and
+  // core/support after that. The role only segments the bracketed sets.
+  const suffix = bracket ? (role ? `:${bracket}:${role}` : `:${bracket}`) : "";
```
howardchung: I would prefer appending bracket and role if they exist to the empty string, rather than nesting like this

geracosta: done. folded #2985 in here too since it's the same line, rather than landing two changes that would conflict.

suffix is built by appending now, and the bracket and role are independent slots, so either works on its own: `...:<hero>::5`, `...:<hero>::core`, `...:<hero>::5:core`. no empty padding needed, since a bracket is 1-8 and a role is text they can't be confused in the same slot.

one thing that came out of making them independent: roles were only being computed for matches that had a bracket, which would have made `?role=core` quietly cover fewer matches than the global set. it's every non-turbo match now, same population as global.

svc/util/queries.ts
```diff
@@ -108,11 +108,22 @@ export async function getHeroItemPopularity(heroId: string) {
...
+  heroId: string,
+  bracket?: number | null,
+  role?: string | null,
+) {
   const ret: AnyDict = {};
   const arr = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.95, 0.99];
-  // Bracketed sets add the 1-8 rank bracket after the global/turbo slot
-  const suffix = bracket ? `:${bracket}` : "";
+  // The bracket and the role are independent slots after the global/turbo one,
+  // so either can be used on its own: ...:<hero>::5, ...:<hero>::core
```
howardchung: is this comment correct? The 5 and the core should be in different slots, right?

so it could be:
:<hero>:5:
:<hero>:5:core
:<hero>::core

I think this is different from the implementation below, where it's storing a key that can have either bracket or role in the same slot. It might be less confusing to always have the same number of slots, with empty data for the "undefined" case.

I would keep the original keys as <hero> with no suffix just for backward compatibility but for all the new filter options, always include the same number of colons/segments

geracosta: you're right. i read your earlier note about appending as being about the key layout too, so i'd collapsed the bracket and the role into one slot.

fixed slots now, with the unfiltered key left alone:

```
...:<hero>:          no filters, unchanged
...:<hero>::5:       bracket
...:<hero>:::core    role
...:<hero>::5:core   both
```

every filtered key carries both segments with an empty one for the filter not in use, so they all have the same number of colons.

this also changes the bracket-only key from `...:<hero>::5` to `...:<hero>::5:`, so the bracket sets from #2972 that are live right now stop being read once this ships. they refill within one retention epoch and the read already falls back to the current epoch, so it's the same sparse window as when brackets first shipped. just didn't want it to look like only the new role sets start empty.

